PR #45, opened 2018-12-13
Bump version and make required fields default

posix4e:


todicus:
NOTE: I used version 0.3.1 to update the allowable task types in #47

PR #48, opened 2019-01-14
Create tests for the escrow smart contract

### LINKED ISSUE #7: Add a comprehensive testsuite
We have currently one test that in a way tests the whole flow. We should decouple the tests into few phases dictated by the different statuses the contract be in.

jjanii:
Closes #7


brdlyptrs:
Looks good to me, dictation and code makes sense.

rbval:
@gaieges they actually are in separate containers: python one in `api` and the node one in `contracts` container.

gaieges:
@rbval sorry meant different images

contracts/Escrow.sol
```diff
@@ -0,0 +1,275 @@
...
+    uint256 private recordingOracleStake;
+
+    address private canceler;
+    address private eip20;
+
+    string private manifestUrl;
+    string private manifestHash;
+
+    string private intermediateManifestUrl;
+    string private intermediateManifestHash;
```
e271828-: Shouldn't these be intermediateResultsUrl and intermediateResultsHash?

jjanii: The contract itself is just a copypaste of the old `hmt.sol` file. I can change these to the ones you suggested if that's what they are supposed to be.

posix4e: No let's keep to the copy/paste approach

PR #63, opened 2019-02-21
set default requester_restricted_answer_set for image_label_area_select

gaieges:
Related to this comment: https://github.com/hCaptcha/hcaptcha-v1/issues/231#issuecomment-466197877

@e271828- @todicus lmk if you think this is reasonable

### DIFF
--- a/basemodels/__init__.py
+++ b/basemodels/__init__.py
@@ -33,6 +33,15 @@ class Manifest(Model):
     job_total_tasks = IntType()
 
     requester_restricted_answer_set = DictType(DictType(StringType))
+
+    def validate_requester_restricted_answer_set(self, data, value):
+        """image_label_area_select should always have a single RAS set"""
+        if data['request_type'] == 'image_label_area_select':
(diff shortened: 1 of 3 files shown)


basemodels/__init__.py
```diff
@@ -33,6 +33,15 @@ class Manifest(Model):
     job_total_tasks = IntType()
 
     requester_restricted_answer_set = DictType(DictType(StringType))
+
+    def validate_requester_restricted_answer_set(self, data, value):
```
todicus: I don't see where this function is called; automatically on manifest.validate()?

gaieges: Yep! That occurs in the tests and in the job creation in exchange

todicus:
Seems fine; if a Requester does not include a RAS, we'll create it for them with answer key "label" and value an empty translation dict. The final results will thus contain a `class_attributes` with "label" as a primary key in the `task_summary`. If the Requester put the actual name of the object to be selected, such as "GATE", then "GATE would be the primary key in the task_summary.

PR #81, opened 2019-03-13
Bump lock file

gaieges:
Replaces Alex's PR because travis was being a punk


gaieges:
Reviewed already, merging

PR #82, opened 2019-03-19
Add some debugging

posix4e:


gaieges:
Should we merge?  You'll need to bump the setup.py version, and once merged tag the release to get the free pypi build/push

PR #85, opened 2019-03-25
add get_pk_from_address to eth_bridge

### LINKED ISSUE #86: Add support for public key lookup
We are already seeing copy paste in many of our services as `get_pk_from_address`. Let's move that to be used from the hmt-escrow itself.

Definition of Done:
1. Add `get_pk_from_address` to `eth_bridge`
2. Setup `eth-kvstore` to the local blockchain infra like we do `hmt-token`: https://github.com/hCaptcha/hmt-escrow/blob/master/docker-compose.yml#L37
3. Add doctests for setting a new pk, and fetching that pk with the same address that stored it.

gaieges:
This logic should live here instead of hmt-servers or loader.

- [x] Still need doctests before merge
- [x] Needs kvstore deployed to local blockchain in order to finish the above

Note: renames `get_pk_from_address` to `get_pub_key_from_addr` and adds a corresponding SET function: `set_pub_key_at_addr`

Fixes #86

### DIFF
--- a/bin/wait_then_run
+++ b/bin/wait_then_run
@@ -2,15 +2,15 @@
 set -exu
 
 KVSTORE_ADDRESS_FILE=/deployed-kvstore/ethkvstore.address.json
-HMTTOKEN_ADDRESS_FILE=/deployed-hmttoken/hmt.address.json
+HMTOKEN_ADDR_FILE=/deployed-hmtoken/hmt.address.json
 
 
-until test -f $HMTTOKEN_ADDRESS_FILE; do
-  echo "waiting for hmt token file at $HMTTOKEN_ADDRESS_FILE .."
+until test -f $HMTOKEN_ADDR_FILE; do
+  echo "waiting for hmt token file at $HMTOKEN_ADDR_FILE .."
   sleep 2
 done
 
-export HMTOKEN_ADDR=$(cat $HMTTOKEN_ADDRESS_FILE | jq --raw-output '.address')
+export HMTOKEN_ADDR=$(cat $HMTOKEN_ADDR_FILE | jq --raw-output '.address')
 
 until test -f $KVSTORE_ADDRESS_FILE; do
   echo "waiting for $KVSTORE_ADDRESS_FILE .."
--- a/bin/wait_then_run
+++ b/bin/wait_then_run
@@ -1,12 +1,9 @@
 #!/bin/bash
 set -exu
-DONE=false
-JOB=${1-'python3 hmt_escrow/job.py -v'}
-STORAGE=${1-'python3 hmt_escrow/storage.py -v'}
-ETH_BRIDGE=${1-'python3 hmt_escrow/eth_bridge.py -v'}
 
 KVSTORE_ADDRESS_FILE=/deployed-kvstore/ethkvstore.address.json
 HMTTOKEN_ADDRESS_FILE=/deployed-hmttoken/hmt.address.json
+
 
 until test -f $HMTTOKEN_ADDRESS_FILE; do
   echo "waiting for hmt token file at $HMTTOKEN_ADDRESS_FILE .."
@@ -27,6 +24,12 @@ until curl --silent ipfs:5001; do
   sleep 1
 done
 
-$JOB
-$STORAGE
-$ETH_BRIDGE
+
+export PYTHONPATH=$(pwd)
+if [ ! -z $* ]; then
+  $*
+else
+  python3 hmt_escrow/job.py -v
+  python3 hmt_escrow/storage.py -v
+  python3 hmt_escrow/eth_bridge.py
+fi
--- a/bin/wait_then_run
+++ b/bin/wait_then_run
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -exu
 
-KVSTORE_ADDRESS_FILE=/deployed-kvstore/ethkvstore.address.json
+KVSTORE_ADDR_FILE=/deployed-kvstore/ethkvstore.address.json
 HMTOKEN_ADDR_FILE=/deployed-hmtoken/hmt.address.json
 
 
@@ -12,12 +12,12 @@ done
 
 export HMTOKEN_ADDR=$(cat $HMTOKEN_ADDR_FILE | jq --raw-output '.address')
 
-until test -f $KVSTORE_ADDRESS_FILE; do
-  echo "waiting for $KVSTORE_ADDRESS_FILE .."
+until test -f $KVSTORE_ADDR_FILE; do
+  echo "waiting for $KVSTORE_ADDR_FILE .."
   sleep 2
 done
 
-export KVSTORE_CONTRACT=$(cat $KVSTORE_ADDRESS_FILE | jq --raw-output '.address')
+export KVSTORE_CONTRACT=$(cat $KVSTORE_ADDR_FILE | jq --raw-output '.address')
 
 until curl --silent ipfs:5001; do
   echo "sleeping"
--- a/bin/wait_then_run
+++ b/bin/wait_then_run
@@ -5,14 +5,24 @@ JOB=${1-'python3 hmt_escrow/job.py -v'}
 STORAGE=${1-'python3 hmt_escrow/storage.py -v'}
 ETH_BRIDGE=${1-'python3 hmt_escrow/eth_bridge.py -v'}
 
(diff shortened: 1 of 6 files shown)


gaieges:
Welp .. @rbval and I realized that we will need a method to deploy the kvstore contract into the local blockchain in order to properly test this via the doctest (similar to the way we do with hmt-token) .. so we're blocked on that..

gaieges:
Yep!

gaieges:
Should be good now, I'm sure a test will break that i'll fix

hmt_escrow/eth_bridge.py
```diff
@@ -232,6 +238,46 @@ def deploy_factory(gas: int = GAS_LIMIT, **credentials) -> str:
...
+    >>> import os
+    >>> from web3 import Web3
+    >>> get_pk_from_address('blah')
+    Traceback (most recent call last):
+    File "<stdin>", line 1, in <module>
+    File "/work/hmt_escrow/eth_bridge.py", line 255, in get_pk_from_address
+        raise ValueError('environment variable GAS_PAYER required')
+    ValueError: environment variable GAS_PAYER required
+    >>> os.environ['GAS_PAYER'] = Web3.toChecksumAddress('0x3895d913a9a231d2b215f402c528511b569c676d')
+    >>> get_pk_from_address(Web3.toChecksumAddress('0xc23546d9c10e85322cb63112ec034090b407a019'))
```
rbval: This doesn't really test that we fetched the correct public key.

We should probably add a "set" method as well so we can test that function actually gives us the correct public key.

Also would rename it from `get_pk_from_address` to `get_pub_key_from_addr`.

hmt_escrow/eth_bridge.py
```diff
@@ -232,6 +238,46 @@ def deploy_factory(gas: int = GAS_LIMIT, **credentials) -> str:
...
+    # TODO: Should we try to get the checksum address here instead of assuming user will do that?
+    GAS_PAYER = os.getenv('GAS_PAYER')
+
+    if not GAS_PAYER:
+        raise ValueError('environment variable GAS_PAYER required')
+
+    w3 = get_w3()
+
+    kvstore = w3.eth.contract(address=KVSTORE_CONTRACT, abi=kvstore_abi)
+    address_pk = kvstore.functions.get(GAS_PAYER, wallet_addr).call({
```
rbval: Go from `pk` to `pub_key`

hmt_escrow/eth_bridge.py
```diff
@@ -232,6 +238,46 @@ def deploy_factory(gas: int = GAS_LIMIT, **credentials) -> str:
...
+        'from':
+        GAS_PAYER
+    })
+    bytes_address = bytes(address_pk, encoding='utf-8')
```
rbval: `bytes_address` to `bytes_addr`

docker-compose.yml
```diff
@@ -13,13 +13,15 @@ services:
...
     image: hmtcentral.azurecr.io/contracts:latest
-    links: 
+    links:
       - ganache
       - ipfs
       - hmt
     volumes:
-      - hmt_token_address:/deployed
-  
+      - hmt_token_address:/deployed-hmttoken
```
rbval: I'd like to avoid double t. `deployed_hmtoken` is prettier

bin/wait_then_run
```diff
@@ -5,14 +5,24 @@ JOB=${1-'python3 hmt_escrow/job.py -v'}
 STORAGE=${1-'python3 hmt_escrow/storage.py -v'}
 ETH_BRIDGE=${1-'python3 hmt_escrow/eth_bridge.py -v'}
 
-until test -f /deployed/hmt.address.json; do
-  echo "waiting for hmt.address.json .."
+KVSTORE_ADDRESS_FILE=/deployed-kvstore/ethkvstore.address.json
+HMTTOKEN_ADDRESS_FILE=/deployed-hmttoken/hmt.address.json
```
rbval: avoid double t. `/deployed-hmtoken/hmt.address.json` is prettier

bin/wait_then_run
```diff
@@ -5,14 +5,24 @@ JOB=${1-'python3 hmt_escrow/job.py -v'}
...
+
+until test -f $HMTTOKEN_ADDRESS_FILE; do
```
rbval: avoid double t and address. `$HMTOKEN_ADDR_FILE` is prettier.

hmt_escrow/eth_bridge.py
gaieges: Ya I can do those renames .. Also this was from the other day when the tests didn't work, fixing them up now.  Was just trying to figure out how to do a test set of a kv pair, but I don't think testing that or the function should live in this repo .. we should use the api of kvstore

hmt_escrow/eth_bridge.py
gaieges: sounds good

docker-compose.yml
gaieges: ok

gaieges:
Will fix up those renames .. lemme get this test working in the current form then I'll address the renames

hmt_escrow/eth_bridge.py
rbval: Well our kvstore returns a string when we do the get. `get_pub_key_from_addr` converts it to bytes so we add a bit functionality there which would be good to test. Can't count the amount of times and hours wasted because I was expecting bytes or str and vice versa.

hmt_escrow/eth_bridge.py
```diff
@@ -232,6 +238,100 @@ def deploy_factory(gas: int = GAS_LIMIT, **credentials) -> str:
...
+      File "/usr/lib/python3.6/doctest.py", line 1330, in __run
+        compileflags, 1), test.globs)
+      File "<doctest __main__.get_pub_key_from_addr[2]>", line 1, in <module>
+        get_pub_key_from_addr('blah')
+      File "hmt_escrow/eth_bridge.py", line 268, in get_pub_key_from_addr
+        raise ValueError('environment variable GAS_PAYER required')
+    ValueError: environment variable GAS_PAYER required
+    >>> os.environ['GAS_PAYER'] = "0x1413862C2B7054CDbfdc181B83962CB0FC11fD92"
+    >>> os.environ['GAS_PAYER_PRIV'] = "28e516f1e2f99e96a48a23cea1f94ee5f073403a1c68e818263f0eb898f1c8e5"
+    >>> set_pub_key_at_addr('blah')  #doctest: +ELLIPSIS
```
rbval: Let's go with `b"2dbc2c2c86052702e7c219339514b2e8bd4687ba1236c478ad41b43330b08488c12c8c1797aa181f3a4596a1bd8a0c18344ea44d6655f61fa73e56e743f79e0d"` instead of `'blah'` as this is documentation at the same time.

gaieges:
> Last time I removed this I had problems in the consuming end

Removed what? the whole function? 

>  I'd probably test in bcs with this branches commit hash just to be sure.

Not a reason to hold this up, lets get it merged and we can pull it in

hmt_escrow/eth_bridge.py
gaieges: done

hmt_escrow/eth_bridge.py
```diff
@@ -232,6 +238,100 @@ def deploy_factory(gas: int = GAS_LIMIT, **credentials) -> str:
...
+
+
+    >>> from web3 import Web3
+    >>> import os
+    >>> os.environ['GAS_PAYER'] = "0x1413862C2B7054CDbfdc181B83962CB0FC11fD92"
+    >>> os.environ['GAS_PAYER_PRIV'] = "28e516f1e2f99e96a48a23cea1f94ee5f073403a1c68e818263f0eb898f1c8e5"
+    >>> pub_key_to_set = b"2dbc2c2c86052702e7c219339514b2e8bd4687ba1236c478ad41b43330b08488c12c8c1797aa181f3a4596a1bd8a0c18344ea44d6655f61fa73e56e743f79e0d"
+    >>> set_pub_key_at_addr(pub_key_to_set)  #doctest: +ELLIPSIS
+    AttributeDict({'transactionHash': ...})
+    """
```
rbval: Nitpicking before perfect: one space needs to be added before closing docstring.

gaieges: Done but we're not aiming for perfection here folks ..

PR #87, opened 2019-03-26
Fix timeout issues regarding IFPS and manifest polling

### LINKED ISSUE #84: Separate ipfs connection to its own function
We experience the urllib3 connect reuse every time we instantiate a new Job which results in an error:
```
DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): ipfs:5001
DEBUG:urllib3.connectionpool:http://ipfs:5001 "GET /api/v0/cat?stream-channels=true&arg= HTTP/1.1" 500 None
```
we could fix this by letting the client using hmt-escrow store a variable of the ipfs client that they would use to pass to the `download` and `upload` instead.

rbval:
Closes #84 

This PR:
1. Makes the IPFS connection a bit neater with timeout decorators
2. Encapsulates helper methods inside Job class as they are used only from there
3. Adds support for polling of the manifest url and hash of a recently deployed Job (useful for bcs)


hmt_escrow/storage.py
```diff
@@ -1,31 +1,39 @@
...
+
+@timeout_decorator.timeout(20)
+def _connect(host: str, port: int) -> Client:
     try:
-        API = ipfsapi.connect(_host, _port)
+        IPFS_CLIENT = ipfsapi.connect(host, port)
+        return IPFS_CLIENT
     except Exception as e:
         raise e
         LOG.error("Connection with IPFS failed because of: {}".format(e))
```
posix4e: Why would this logger ever fire?

rbval: Should be fixed now, that what you meant?

gaieges: He's saying you'll never hit this line, might as well remove it or move it before the raise

rbval: already moved it, wanted to make sure I understood it correctly

gaieges: Oh i see, I was looking at the filtered commit.  I don't see the raise anymore

rbval: here: https://github.com/hCaptcha/hmt-escrow/blob/f2b88775cad5eaa5bfcca1e4011dc020949f1ec7/hmt_escrow/storage.py#L29

gaieges: lgtm

PR #89, opened 2019-04-02
Add public getters

rbval:
Adds some convenience functions for third parties that don't require the initialization of the Job class:
- `status`
- `manifest_url`
- `manifest_hash`

### DIFF
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -25,6 +25,105 @@ LOG = logging.getLogger("hmt_escrow.job")
 Status = Enum('Status', 'Launched Pending Partial Paid Complete Cancelled')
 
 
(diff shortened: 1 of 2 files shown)


hmt_escrow/job.py
```diff
@@ -25,6 +25,105 @@
...
+        
+    Returns:
+        Enum: returns the status as an enumeration.
+
+    """
+    status_ = escrow_contract.functions.getStatus().call({
+        'from': gas_payer,
+        'gas': gas
+    })
+    return Status(status_ + 1)
```
gaieges: `+1`? For backward compatibility?

rbval: yap!

PR #91, opened 2019-04-09
Add public intermediates

rbval:
Add intermediate results and hash as public getter functions.

### DIFF
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -308,10 +308,6 @@ class Job:
 
         # Access an existing Job.
         elif escrow_addr and factory_addr and not escrow_manifest:
-            if not self._factory_contains_escrow(escrow_addr, factory_addr):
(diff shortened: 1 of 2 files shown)


hmt_escrow/job.py
```diff
@@ -124,6 +124,50 @@ def manifest_hash(escrow_contract: Contract,
...
+    """Retrieves the deployed intermediate results url uploaded on Job initialization.
+
+    Args:
+        escrow_contract (Contract): the escrow contract of the Job.
+        gas_payer (str): an ethereum address paying for the gas costs.
+        gas (int): maximum amount of gas the caller is ready to pay.
+    
+    Returns:
+        str: returns the intermediate results url of Job's escrow contract.
+
```
gaieges: want a doctest here?  (might not be worth it for something this small, just provoking the idea)

PR #93, opened 2019-04-25
bump Python to 3.7, update packages

e271828-:

PR #95, opened 2019-04-26
Update/basemodels

Villux:
Same as for basemodels repo.

Yet again command `pipenv update hmt-basemodels` updated other deps as well. What might be the reason for this?


Villux:
Will conflict with https://github.com/hCaptcha/hmt-escrow/pull/93 most likely. Maybe we can merge these two together?

PR #116, opened 2019-10-21
Require address

posix4e:
Temp workaround to restrict who can launch contracts.


posix4e:
Cancelling this @e271828- has vetoed any change of this type to a public repo

PR #120, opened 2019-11-13
Gas optimize

nickgeoca:
Added gas optimizations. Including
- Change finalAmounts from storage to memory. **1.5M gas savings** per 100 recipients for bulkPayout
- Enable solc optimize. **20k gas savings** for per 100 recipients for bulkPayout
- Changes functions from public to external. **~250 gas savings** for setup function. 

Not included
- Inherit SafeMath instead of delegate call. This makes auditing a little more challenging but might worth it. Estimated to save about ~97k gas per 100 recipients for bulkPayout. 
- bulkPayout public to external. Was getting stack too deep errors. But likely around ~2.5k gas savings per 100 recipients for bulkPayout

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,6 +103,10 @@ ENV/
 # vim
 *.swp
 
+# emacs
+*~
+*[#]*
+
 .idea
 venv*
 het/secrets.json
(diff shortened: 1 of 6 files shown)


nickgeoca:
Fixed! Will do.

nickgeoca:
Is the Travis Client failing because the devnet port was set to 8545?

Issue #79, opened 2019-03-12
Append  results


posix4e:
We should have some way of automatically appending to the  results.

```
1. Generate a ipns links (intermediate and final) on startup of the contract both results When you launch the contract  https://github.com/hCaptcha/hmt-escrow/blob/master/hmt_escrow/job.py#L348
2. Update  results ipfs and add new ipns to new ipfs at https://github.com/hCaptcha/hmt-escrow/blob/master/hmt_escrow/job.py#L469
3. Update results with new ipfs and update ipns
https://github.com/hCaptcha/hmt-escrow/blob/master/hmt_escrow/job.py#L637
4. https://github.com/hCaptcha/hmt-escrow/blob/master/hmt_escrow/job.py#L652
Have complete download newest ipfs results, and add hash.
```

posix4e:
- So it turns out you can use ipns to just automatically update the pointer in the contract. This means we only need to update the intermediate results ipns endpoint.
- We don't need to update the hash until we are done with contract completely.

nickgeoca:
The proposed design is to give each Escrow a unique IPNS link. This can be done by generating an IPFS keypair for each IPNS link. The end result of an Escrow's IPNS link will be: an IPFS with all Escrow results/manifest/etc data in it.

**changes to the smart contract**
1) Put the IPNS link onchain during Escrow construction.
2) Add an option to update the hashes (when needed) in the `storeResults` and `bulkPayOut` code. For example, `bulkPayOut` can be called without writing the final results URL & hash.

**changes to the python** (1 & 2 are the same as above)
 1) Create an IPNS link then pass it to Escrow on construction. 
 2) Optional write results/manifest/etc data onchain.
 3) Our python `upload` function puts results/manifest/etc data on IPFS. The changes to `upload` are this: a) `upload` will concatenate the results/manifest/etc data to the data of all previous uploads for a given escrow; b) load this new data file to IPFS; c) set the IPNS url to point to the new IPFS data file.

nickgeoca:
Maybe putting cart before the horse, but here are some open questions:
1) What is a good way to serialize/deserialize the results/manifest/etc data, given it is a series of concatenations? Something along the lines of? ... `<manifest-data-1><delimiter><manifest-data-2><delimiter>` . What is a good binary delimiter?
[...]

PR #122, opened 2019-12-11
Switch to storing results in a static IPNS link

nickgeoca:
Use ipns. Rename variables in Escrow.sol


test/Escrow.js
```diff
@@ -310,6 +314,34 @@ contract('Escrow', (accounts) => {
...
+
+        const amountToPay = [4];
+        assert(await Escrow.getReputationOracleResultsUrl.call() === '' && await Escrow.getReputationOracleResultsHash.call() === '',
+              'hash and url should be empty');
+        await Escrow.bulkPayOut(toAddress, amountToPay, url, hash, '000', { from: reputationOracle });
+        assert(await Escrow.getReputationOracleResultsUrl.call() === url && await Escrow.getReputationOracleResultsHash.call() === hash,
+              'either hash or url not set');
+        await Escrow.bulkPayOut(toAddress, amountToPay, '', '', '000', { from: reputationOracle });
+        assert(await Escrow.getReputationOracleResultsUrl.call() === url && await Escrow.getReputationOracleResultsHash.call() === hash,
+                    'hash or url changed when it should not have');
```
posix4e: where's the test to make sure ipfs updates?

hmt_escrow/job.py
```diff
@@ -459,19 +447,24 @@ def bulk_payout(self,
...
             pub_key (bytes): the public key of the Reputation Oracle.
+            gas (int): Optional, gas limit
+            storeOnchain (bool): Store data onchain. Saves 10k gas
 
         Returns:
             bool: returns True if paying to ethereum addresses and oracles succeeds.
 
         """
-        (hash_, url) = upload(results, pub_key)
+        (hash_, url) = upload(results, pub_key, ipnsKeypairName='reputation-oracle-results-' + self.job_contract.address)
```
gaieges: often useful to do a f type string as its a bit cleaner: `f'reputation-oracle-results-{self.job_contract.address}'`

hmt_escrow/storage.py
```diff
@@ -70,9 +70,49 @@ def download(key: str, private_key: bytes) -> Dict:
...
+
+    Args:
+        name (str): Name we call ipns link
+
+    Returns:
+        str: Returns the IPNS url
+    """
+    keys = IPFS_CLIENT.key.list()
+    try:
+        return 'https://ipfs.io/ipns/' + list(filter(lambda x: x['Name'] == name.lower(), keys['Keys']))[0]['Id']
```
gaieges: do we want to hard code this ipfs url? might be better as an env var.  

Is this the only way users would look up the link address?

hmt_escrow/storage.py
```diff
@@ -70,9 +70,49 @@ def download(key: str, private_key: bytes) -> Dict:
...
+    except Exception as e:
+        return ''
+    return ''
```
gaieges: why not let the exceptions occur here? if theres a problem you don't want to have to write additional code to check for a blank string

contracts/Escrow.sol
```diff
@@ -171,14 +192,14 @@ contract Escrow {
             status == EscrowStatuses.Partial,
             "Escrow not in Pending or Partial status state"
         );
-        intermediateResultsUrl = _url;
-        intermediateResultsHash = _hash;
-        emit IntermediateStorage(_url, _hash);
+        recordingOracleResultsUrl = _url;
+        recordingOracleResultsHash = _hash;
+        emit RecordingOracleResults(_url, _hash);
```
gaieges: I like the rename

contracts/Escrow.sol
```diff
@@ -171,14 +192,14 @@ contract Escrow {
...
     }
 
     function bulkPayOut(
```
rbval: While you're at it, let's rename these to `bulkPayment`. This won't then be confused with the PayPal payouts.

hmt_escrow/job.py
```diff
@@ -121,69 +121,49 @@ def manifest_hash(escrow_contract: Contract,
...
-    return escrow_contract.functions.getIntermediateResultsHash().call({
+    return escrow_contract.functions.getRecordingOracleResultsHash().call({
         'from':
         gas_payer,
         'gas':
         gas
     })
 
-
-def launcher(escrow_contract: Contract, gas_payer: str,
```
rbval: This probably shouldn't be deleted, since we use it to whitelist in another service.

hmt_escrow/job.py
nickgeoca: done

hmt_escrow/job.py
```diff
@@ -243,8 +223,8 @@ def __init__(self,
         True
         >>> job.setup()
         True
-        >>> job.launcher()
-        "0x1413862C2B7054CDbfdc181B83962CB0FC11fD92"
```
rbval: We do want to check that the launcher is the same that launched the contract. I would not remove or modify the existing test.

contracts/Escrow.sol
nickgeoca: cool. @gaieges , we are going to do this in a separate commit actually!

hmt_escrow/job.py
```diff
@@ -459,19 +447,24 @@ def bulk_payout(self,
...
         """
-        (hash_, url) = upload(results, pub_key)
+        (hash_, url) = upload(results, pub_key, ipnsKeypairName='reputation-oracle-results-' + self.job_contract.address)
         eth_addrs = [eth_addr for eth_addr, amount in payouts]
         hmt_amounts = [int(amount * 10**18) for eth_addr, amount in payouts]
 
         txn_func = self.job_contract.functions.bulkPayOut
-        func_args = [eth_addrs, hmt_amounts, url, hash_, 1]
+
+        chainUrl  = url   if storeOnchain else ''
```
rbval: I would modify these camelCases to snake_cases to have a unified way to name variables. More pythonish.

hmt_escrow/storage.py
nickgeoca: The thought was that if there is no key in the list, then you would not want to throw an exception. Maybe we can do a test beforehand? `nameExists(name) -> bool`

hmt_escrow/job.py
```diff
@@ -838,12 +836,33 @@ def final_results(self, priv_key: bytes, gas: int = GAS_LIMIT) -> Dict:
...
             'from': self.gas_payer,
             'gas': gas
         })
-        return download(final_results_url, priv_key)
+        return download(reputation_oracle_results_url, priv_key)
+
+    def getIpnsUrl(self, keyName: str) -> str:
+        return getIpnsLink(keyName)
+
+    def launcher(self, escrow_contract: Contract, gas_payer: str,
```
rbval: Ah, saw you moved it here. Any particular reason why?

hmt_escrow/storage.py
rbval: I would use an env variable.

Also the oneliner is a bit unclear. I would maybe use a couple of lines there myself.

nickgeoca: One example of not using the 'https://...' is publishing the IPNS address offline. I don't know any other counter examples, but it is probably a good idea to add an ENV var.

contracts/Escrow.sol
```diff
@@ -21,17 +24,22 @@ contract Escrow {
     string private manifestUrl;
     string private manifestHash;
 
-    string private intermediateResultsUrl;
-    string private intermediateResultsHash;
+    // aka internal, not customer results
```
rbval: I would remove these comments, public repo shouldn't have indications regarding customers

contracts/Escrow.sol
```diff
@@ -21,17 +24,22 @@ contract Escrow {
...
-    string private finalResultsUrl;
-    string private finalResultsHash;
+    // aka customer results
+    string private reputationOracleResultsUrl;
+    string private reputationOracleResultsHash;
 
     uint private expiration;
 
-    uint256[] private finalAmounts;
+    uint256[] private reputationOraclePayouts;
```
rbval: I'd use payment instead of payout here as well.

contracts/Escrow.sol
```diff
@@ -163,6 +183,7 @@ contract Escrow {
         }
     }
 
+    // recording oracle is the intenral hash/url
```
rbval: a bit unclear comment, little typo as well

contracts/Escrow.sol
```diff
@@ -193,22 +214,26 @@ contract Escrow {
...
+        if (balance < aggregatedBulkPayout) {
             return bulkPaid;
         }
 
-        finalResultsUrl = _url;
-        finalResultsHash = _hash;
+        bool writeOnchain = bytes(_hash).length != 0 || bytes(_url).length != 0;
+        if (writeOnchain) {
+          // Be sure they are both zero if one of them is
+          reputationOracleResultsUrl = _url;
```
rbval: Doesn't `_url` come from the contract fields? Is there a possibility that `_url` might not be empty from last chunk's payments? One contract can facilitate the HMT payments of multiple chunks.

contracts/Escrow.sol
```diff
@@ -193,22 +214,26 @@ contract Escrow {
...
+          reputationOracleResultsHash = _hash;
+        }
 
-        (uint256 reputationOracleFee, uint256 recordingOracleFee) = finalizePayouts(_amounts);
+        (uint256 reputationOracleFee, uint256 recordingOracleFee) = finalizePayouts(_payouts);
         HMTokenInterface token = HMTokenInterface(eip20);
-        if (token.transferBulk(_recipients, finalAmounts, _txId) == _recipients.length) {
-            delete finalAmounts;
+        if (token.transferBulk(_recipients, reputationOraclePayouts, _txId) == _recipients.length) {
+            delete reputationOraclePayouts;
```
rbval: Do we get access to this field in later chunk payments?

hmt_escrow/job.py
```diff
@@ -1090,17 +1110,19 @@ def _create_escrow(self, gas: int = GAS_LIMIT) -> bool:
...
+        func_args = []
         txn_info = {
             "gas_payer": self.gas_payer,
             "gas_payer_priv": self.gas_payer_priv,
             "gas": gas
         }
-
-        handle_transaction(txn_func, *[], **txn_info)
-        return True
+        handle_transaction(txn_func, *func_args, **txn_info)
```
rbval: This change makes sense. This however might be a quite big breaking change in itself. Make sure to check this out in all the doctests as we might be using this quit a lot (`._create_escrow()`)

hmt_escrow/storage.py
```diff
@@ -70,9 +70,49 @@ def download(key: str, private_key: bytes) -> Dict:
...
 
+def createNewIpnsLink(name: str) -> str:
+    """Creates a new IPFS Id. The IPNS links are managed by key value system.
+       Pass in the key (e.g. 'intermediate-results-' + self.job_contract.address), 
+       then creats the ipns link
 
+    >>> import random 
+    >>> keyName = str(random.getrandbits(32 * 8)) # get random, or else throws duplicate key error
+    >>> createNewIpnsLink(keyName) != ''
+    True
```
rbval: Doctests should come after the documentation.

hmt_escrow/storage.py
```diff
@@ -70,9 +70,49 @@ def download(key: str, private_key: bytes) -> Dict:
...
+    key = IPFS_CLIENT.key.gen(name, 'ed25519')
+    return getIpnsLink(name)
+
+def getIpnsLink(name: str) -> str:
+    """Get the ipns link with the name of it which we remember it by
+
+    >>> import random 
+    >>> keyName = str(random.getrandbits(32 * 8)) # get random, or else throws duplicate key error
+    >>> createNewIpnsLink(keyName) != ''
+    True
```
rbval: Here as well, doctests after documentation. Check https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

test/Escrow.js
nickgeoca: 622 and 629 show the updating
https://github.com/hCaptcha/hmt-escrow/blob/bc5a2b230c5f0e6cbf384aabc4185dab0132e25a/hmt_escrow/job.py#L629

the js file is just setting the ipns link. so ipfs updates aren't touched on the blockchain

hmt_escrow/job.py
nickgeoca: k, it is currently in!
https://github.com/hCaptcha/hmt-escrow/blob/ipns/hmt_escrow/job.py#L857

hmt_escrow/job.py
nickgeoca: It was getting a different address than `0x1413862C2B7054CDbfdc181B83962CB0FC11fD92`. Is it supposed to be deterministic? If not maybe we can get the event from the factory.

hmt_escrow/job.py
nickgeoca: No reason. I'll restore the order of functions from before.

nickgeoca:
`def getIpnsLink(name: str) -> str:`
currently name is being converted to lower case. b/c ethereum addresses are mixed case- for error detection purposes. an example is these might both be valid addresses `0xFeef...` & `0xfeEf...`

converting to all lower case fixes this. would we rather convert to all lower case in storage.py, where it is right now, or higher up in the stack?
@rbval

hmt_escrow/storage.py
gaieges: Hm, I see.. maybe `None` is the better response vs an empty string?  Idk my preference here would be to more narrowly do a try/catch and throw an exception if there is something they're doing wrong from the code perspective (ie if no name found, then `raise ValueError('no name found')`. 

If the code and usage of the api is correct, and a None/empty string is a valid response that can be used just like a real url, then empty string might be right.

gaieges: IMO it's better to be explicit about why something fails, and throw a specific exception so that higher up we can differentiate where the problem is - ie is it a user input failure or is it a code failure that we might want to handle different ways

hmt_escrow/job.py
gaieges: The order doesn't matter much, just harder to see / not as obvious in a diff view when reviewing

hmt_escrow/job.py
```diff
@@ -1071,6 +1098,11 @@ def _last_escrow_addr(self, gas: int = GAS_LIMIT) -> str:
             gas
         })
 
+# goes at end of _create_escrow test
+#        >>> get_ipns_link(addr)
```
rbval: These comments should be removed after moving

hmt_escrow/storage.py
```diff
@@ -70,9 +71,79 @@ def download(key: str, private_key: bytes) -> Dict:
     msg = _decrypt(private_key, ciphertext)
     return json.loads(msg)
 
+def create_new_ipns_link(name: str) -> str:
+    """Creates a new IPFS Id. The IPNS links are managed by key value system.
+       Pass in the key (e.g. f'intermediate-results-{self.job_contract.address}'), 
+       then creats the ipns link
 
+    >>> import random 
+    >>> keyName = str(random.getrandbits(32 * 8)) # get random, or else throws duplicate key error
```
rbval: here camelCase as well

hmt_escrow/storage.py
```diff
@@ -91,28 +162,39 @@ def upload(msg: Dict, public_key: bytes) -> Tuple[str, str]:
...
         manifest_ = json.dumps(msg, sort_keys=True)
     except Exception as e:
         LOG.error("Can't extract the json from the dict")
         raise e
-
+    
     hash_ = hashlib.sha1(manifest_.encode('utf-8')).hexdigest()
     try:
-        key = IPFS_CLIENT.add_bytes(_encrypt(public_key, manifest_))
+        ipfsFileHash = IPFS_CLIENT.add_bytes(_encrypt(public_key, manifest_))
```
rbval: camelCase

docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
       - GAS_LIMIT=6700000
       - IPFS_HOST=ipfs
       - IPFS_PORT=5001
+      - IPNS_PATH=https://ipfs.io/ipns/
```
rbval: Could we locally use the local ipfs? Shouldn't our local ipfs support ipns as well? Also, our tests seem to be using the ipns, probably good that our tests don't make calls to https://ipfs.io/ipns

test/Escrow.js
nickgeoca: @posix4e , this was just added in hmt_escrow/test_ipns.py 
ill mark resolved

hmt_escrow/job.py
gaieges: I don't think deterministic .. @rbval ?

hmt_escrow/job.py
nickgeoca: Ok, it has been restored. If there was a reason i don't remember, but it makes sense for it to be up there.

hmt_escrow/job.py
nickgeoca: Oh, nvm, ur comment makes sense

hmt_escrow/job.py
nickgeoca: Let's stick w/ previous style for now. Race conditions on getting the last address could cause problems

bin/wait_then_run
```diff
@@ -32,4 +32,5 @@ else
   python3 hmt_escrow/job.py
   python3 hmt_escrow/storage.py
   python3 hmt_escrow/eth_bridge.py
+  python3 hmt_escrow/test_ipns.py
```
gaieges: Can you pull this into a `tests` directory rather than the library directory?

docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
       - GAS_LIMIT=6700000
       - IPFS_HOST=ipfs
       - IPFS_PORT=5001
+      - IPNS_PATH=ipfs:IPFS_PORT/ipns/
```
gaieges: Does this (port) work? I can't imagine it does

hmt_escrow/storage.py
```diff
@@ -107,16 +114,108 @@ def upload(msg: Dict, public_key: bytes) -> Tuple[str, str]:
 
     hash_ = hashlib.sha1(manifest_.encode('utf-8')).hexdigest()
     try:
-        key = IPFS_CLIENT.add_bytes(_encrypt(public_key, manifest_))
+        ipfsFileHash = IPFS_CLIENT.add_bytes(_encrypt(public_key, manifest_))
```
gaieges: var names in python are `snake_cased` not `camelCased`

hmt_escrow/storage.py
```diff
@@ -107,16 +114,108 @@ def upload(msg: Dict, public_key: bytes) -> Tuple[str, str]:
...
     except Exception as e:
         LOG.warning("Adding bytes with IPFS failed because of: {}".format(e))
         raise e
-    return hash_, key
+
+    if ipns_keypair_name != '':
+        try:
+            # publish ipns ... docs: https://ipfs.io/ipns/12D3KooWEqnTdgqHnkkwarSrJjeMP2ZJiADWLYADaNvUb6SQNyPF/docs/http_client_ref.html#ipfshttpclient.Client.name
+            # TODO: is it faster if
```
gaieges: ?

hmt_escrow/storage.py
```diff
@@ -58,25 +47,53 @@ def download(key: str, private_key: bytes) -> Dict:
...
 
 @timeout_decorator.timeout(20)
-def upload(msg: Dict, public_key: bytes) -> Tuple[str, str]:
+def upload(msg: Dict, public_key: bytes,
+           ipns_keypair_name: str = '') -> Tuple[str, str]:
     """Upload and encrypt a string for later retrieval.
     This can be manifest files, results, or anything that's been already
     encrypted.
 
+    Args:
```
gaieges: Might also want to add another note here that describes what it does with IPNS: ie "additionally creates a IPNS link to the uploaded file hash that gets updated as uploads change" (if thats accurate)

hmt_escrow/test_ipns.py
```diff
@@ -0,0 +1,134 @@
...
+import random
+from basemodels import Manifest
+
+from storage import download, upload, get_ipns_link, create_new_ipns_link, _connect, IPFS_CLIENT, ipns_link_exists
+from job import Job
+
+import pdb
+
+# logging = print
+logging = lambda *argv: None
```
gaieges: whats this?

hmt_escrow/test_ipns.py
```diff
@@ -0,0 +1,134 @@
+import unittest
+from unittest.mock import patch
+import random
+from basemodels import Manifest
+
+from storage import download, upload, get_ipns_link, create_new_ipns_link, _connect, IPFS_CLIENT, ipns_link_exists
+from job import Job
+
+import pdb
```
gaieges: needed?

test/Escrow.js
```diff
@@ -310,6 +316,34 @@ contract('Escrow', (accounts) => {
...
+        assert(await Escrow.getFinalResultsUrl.call() === '' && await Escrow.getFinalResultsHash.call() === '',
+              'hash and url should be empty');
+        await Escrow.bulkPayOut(toAddress, amountToPay, url, hash, '000', { from: reputationOracle });
+        assert(await Escrow.getFinalResultsUrl.call() === url && await Escrow.getFinalResultsHash.call() === hash,
+              'either hash or url not set');
+        await Escrow.bulkPayOut(toAddress, amountToPay, '', '', '000', { from: reputationOracle });
+        assert(await Escrow.getFinalResultsUrl.call() === url && await Escrow.getFinalResultsHash.call() === hash,
+                    'hash or url changed when it should not have');
+      } catch (ex) {
+        assert(false);
```
gaieges: does this even need to be caught?

hmt_escrow/storage.py
nickgeoca: this is an api thing.. so right now `upload` does this:
 `upload(msg: Dict, public_key: bytes, ipns_keypair_name: str = '')`
 1) put file on IPFS
 2) if ipns_keypair_name, then pin to IPNS

here are two different examples to clairfy
* upload to IPFS: `upload(data, public_key, '')`
* upload to IPFS and pin to IPNS `upload(data, public_key, 'escrow1')`

altenatively we can chagne the api to be like this
`ipfs_hash = upload(data, public_key)`
`pin_ipns(ipfs_hash, 'escrow1')`

test/Escrow.js
nickgeoca: no, let's remove this next release. i believe i changed this in an audit branch, we can merge that in at some point

gaieges: Audit stuff wasn't merged separately?

hmt_escrow/storage.py
gaieges: I'm mostly just wondering what your TODO item means?

bin/wait_then_run
```diff
@@ -32,4 +32,5 @@ else
   python3 hmt_escrow/job.py
   python3 hmt_escrow/storage.py
   python3 hmt_escrow/eth_bridge.py
+  python3 hmt_escrow/tests/test_ipns.py
```
rbval: Can this be moved to: https://github.com/hCaptcha/hmt-escrow/blob/master/bin/test?

docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
       - GAS_LIMIT=6700000
       - IPFS_HOST=ipfs
       - IPFS_PORT=5001
+      - IPNS_PATH=ipfs:5001/ipns/
```
rbval: I would remove this locally and just build this wherever this is used like this:
```
IPFS_HOST = os.getenv("IPFS_HOST", "localhost")
IPFS_PORT = int(os.getenv("IPFS_PORT", 5001))
IPNS_PATH = os.getenv("IPNS_PATH", f"{IPFS_HOST}:{IPFS_PORT}/ipns/")
```
That way we can remove this from local `docker-compose.yml`

hmt_escrow/eth_bridge.py
```diff
@@ -1,13 +1,13 @@
 import logging
 import os
 import time
+from typing import Any, Dict, List, Optional, Tuple
 
+from hmt_escrow.kvstore_abi import abi as kvstore_abi
```
rbval: This reordering of variables makes this review hard to understand.

hmt_escrow/job.py
rbval: These comments should be removed.

hmt_escrow/storage.py
```diff
@@ -18,6 +19,7 @@
 LOG = logging.getLogger("hmt_escrow.storage")
 IPFS_HOST = os.getenv("IPFS_HOST", "localhost")
 IPFS_PORT = int(os.getenv("IPFS_PORT", 5001))
+IPNS_PATH = os.getenv("IPNS_PATH", "localhost")
```
rbval: This is fixable with an earlier comment

hmt_escrow/storage.py
```diff
@@ -58,24 +48,53 @@ def download(key: str, private_key: bytes) -> Dict:
...
+    >>> pub_key = b"2dbc2c2c86052702e7c219339514b2e8bd4687ba1236c478ad41b43330b08488c12c8c1797aa181f3a4596a1bd8a0c18344ea44d6655f61fa73e56e743f79e0d"
+    >>> job = Job(credentials=credentials, escrow_manifest=manifest)
+    >>> (hash_, manifest_url) = upload(job.serialized_manifest, pub_key)
+    >>> manifest_dict = download(manifest_url, job.gas_payer_priv)
+    >>> manifest_dict == job.serialized_manifest
+    True
     """
+
+    # Q prefix is ipfs. else assume ipns
+    if key[0] != 'Q':
```
rbval: Where did this `Q` come from?

hmt_escrow/storage.py
rbval: Yea, should remove the TODO comment or finish the thought

test/Escrow.js
nickgeoca: No. We wanted to hold off on that if I recall.

hmt_escrow/storage.py
nickgeoca: ah, removed.

hmt_escrow/eth_bridge.py
nickgeoca: this might have been isort

hmt_escrow/storage.py
nickgeoca: we want to see if it is an ipns id. https://github.com/ipfs/faq/issues/22
all ipfs hashes are prefixed with Q. i updated the code for clarity

hmt_escrow/eth_bridge.py
rbval: hmm, that's actually good. let's keep it that way.

gaieges:
Oh @rbval raises a good point around bumping the version .. don't forget to also TAG your git commit once merged to master so that it gets deployment to PyPi

gaieges:
BTW @rbval @nickgeoca  this seems more of a "minor" update than a "patch" update .. I'd propose `0.6.0`

rbval:
True, let's do 0.6.0

nickgeoca:
closes #79

PR #126, opened 2020-01-06
Set results url one time

nickgeoca:
Changes below
- onchain results url only set once
- add a missing fn `job.final_results_url`

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -4,6 +4,3 @@ docker-compose build
 docker-compose run job bash -c "./bin/wait_then_run $*"
 docker-compose run job bash -c "./bin/unittests $*"
 docker-compose run contracts
-
-
-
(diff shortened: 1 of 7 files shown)


nickgeoca:
> 2 things here - I don't see much additional testing here, is this change covered already via other tests?
> 
> Other issue is - I'm not sure how much we use these calls that are being modified (ie from BCS or loader, JR, etc); is there any risk that we're changing something that we currently use?

Truffle test covers the new changes. I don't think too much changes in the python code.

I checked various repos. jobrunner changes. Here is the update to job runner. Otherwise I don't think too much changes.
```
-from hmt_escrow.job import Job, intermediate_url, intermediate_hash, status
+from hmt_escrow.job import Job, intermediate_results_url, intermediate_results_hash, status
```

gaieges:
Ahh I see.  Given that this is a public repo (but still immature API) there should probably be a note somewhere in like a "release notes" doc that notes that change (and any others that might be used)

gaieges:
@rbval can approve

PR #130, opened 2020-01-31
fix: update tests to check error messages

PaddyMc:
### Description
In the current implementation of some tests, when checking if the contract errors some errors are checked incorrectly. This branch fixes that logic and checks specific error messages

- Abstracted the error logic into a util
- Updated test to fail correctly
- Update formatting of tests to be more readable

closes #130

### DIFF
--- a/test/Escrow.js
+++ b/test/Escrow.js
@@ -34,7 +34,7 @@ contract('Escrow', accounts => {
   describe('initialization of Escrow contract works correctly', () => {
     it('status is launched', async () => {
       const initialStatus = await Escrow.getStatus();
-      assert.equal(0, initialStatus);
+      assert.equal(initialStatus, 0);
     });
 
     it('balance is 0', async () => {
@@ -192,7 +192,7 @@ contract('Escrow', accounts => {
         { from: accounts[0] }
       );
       const status = await Escrow.getStatus();
-      assert.equal(1, status);
+      assert.equal(status, 1);
     });
   });
 
--- a/test/Escrow.js
+++ b/test/Escrow.js
@@ -205,10 +205,6 @@ contract('Escrow', accounts => {
         'Address calling not the reputation oracle'
       );
     });
-
-    it('fails if escrow status is not complete or paid', async () => {
-      await testWillThrow(Escrow.complete, [], { from: reputationOracle }, '');
-    });
   });
 
   describe('calling storeResults', () => {
--- a/test/Escrow.js
+++ b/test/Escrow.js
@@ -1,5 +1,6 @@
 const EscrowAbstraction = artifacts.require('Escrow');
 const HMTokenAbstraction = artifacts.require('HMToken');
+const { testWillThrow } = require('./utils/helpers');
 
 let Escrow;
 let HMT;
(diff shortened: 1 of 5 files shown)


rbval:
Hi @PaddyMc  and thanks for the PR. Want to add an issue first about your findings so I can add it to the project board? :)

PaddyMc:
Hi @rbval, sure, created this issue. https://github.com/hCaptcha/hmt-escrow/issues/133 :muscle:

PaddyMc:
@posix4e 
> it'd be nice to break something like this into multiple commits so we can breakup the refactoring from the fix 

For sure will next time :+1:

> What tests assure this change?

Not sure I understand here. Could you elaborate?

posix4e:
@PaddyMc trying to understand what changes to the testing assure this fix worked.

PaddyMc:
If we compare the invocation of this function:

> https://github.com/hCaptcha/hmt-escrow/pull/130/files#diff-6639d32f4f91562f4eec97bd924b494eL128 ✅ 

To

> https://github.com/hCaptcha/hmt-escrow/pull/130/files#diff-6639d32f4f91562f4eec97bd924b494eL119 ❌ 

We can see that the second function is incorrect because there's an extra parameter being passed in and the test will `assert(true)` on any fail, including incorrect ones.

So I [created a util](https://github.com/hCaptcha/hmt-escrow/pull/130/files#diff-0f3de982ede873a6739762317552c6c2) that uses regex to check if the correct error message is contained in the failure. 

Hope that helps 🤝

rbval:
Hi @PaddyMc , initially looks good, will take a closer look later. Meanwhile, could you go and sign this CLA (Contributor License Agreement) https://app.hellosign.com/s/IGBgHJ4y , in order for us to merge your code?

test/Escrow.js
```diff
@@ -10,59 +11,46 @@ const hash = 'fakehash';
...
+  describe('initialization of Escrow contract works correctly', () => {
     it('status is launched', async () => {
-      try {
-        const initialStatus = await Escrow.getStatus.call();
-        assert.equal(0, initialStatus);
-      } catch (ex) {
-        assert(false);
-      }
+      const initialStatus = await Escrow.getStatus();
+      assert.equal(0, initialStatus);
```
rbval: Switch over to `actual`, `expected`.

`assert.equal(initialStatus, 0)`

test/Escrow.js
```diff
@@ -143,63 +180,85 @@ contract('Escrow', (accounts) => {
...
+        1,
+        1,
+        recOIpns,
+        repOIpns,
+        url,
+        hash,
+        { from: accounts[0] }
+      );
+      const status = await Escrow.getStatus();
       assert.equal(1, status);
```
rbval: Here as well to `actual`, `expected`

test/utils/helpers.js
```diff
@@ -0,0 +1,14 @@
...
+		assert(false, 'contract failed to throw an error')
+	} catch (error) {
+		assert(
+			RegExp(errorMessage).test(error),
+			`the error message => ${errorMessage} => does not match => ${error}`
+		)
+	}
+}
+
+module.exports = { testWillThrow }
```
rbval: Minor, should have a newline. Could you add this requirement to the linter as well?

test/Escrow.js
PaddyMc: Updated in [latest commit](https://github.com/hCaptcha/hmt-escrow/pull/130/commits/a276f2f8624e8aeb7be4266173fa33bb70c910dd#diff-6639d32f4f91562f4eec97bd924b494eR37) :handshake:

test/Escrow.js
PaddyMc: Updated in [latest commit](https://github.com/hCaptcha/hmt-escrow/pull/130/commits/a276f2f8624e8aeb7be4266173fa33bb70c910dd#diff-6639d32f4f91562f4eec97bd924b494eR195) :handshake:

test/utils/helpers.js
PaddyMc: Ace, good catch!

Added [eol-last](https://eslint.org/docs/rules/eol-last) to the eslintrc :+1:

Issue #72, opened 2019-03-07
Generate docs from doctests


rbval:
# Description
A lot of our tests are written as doctests (https://docs.python.org/3/library/doctest.html). It would be good to generate documentation from these tests automatically.

Here are few example options to choose from:
- http://www.sphinx-doc.org/en/master/index.html
- https://github.com/BurntSushi/pdoc
- https://github.com/twisted/pydoctor
- http://www.stack.nl/~dimitri/doxygen/index.html

Any other ones can be used as well.

# Todo
1. Choose a python doc generation software
2. Generate docs from our docstrings found in the python packages
3. Update README

posix4e:
agreed. i think someone could simplify publish it to the github page

gaieges:
As I noted to @rbval in DM- I don't find autogen docs to be very useful, but get great amounts of value from the human docs.  

That being said, if we have the doctests, then its a no brainer to add in something like this.

bxabi:
Hi! I started to look at this issue.
Sphinx seems to be widely used, and has support for the doctests. It also can be used to create a readthedocs.io style documentation hosted on that site, which many projects seem to use.
https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html
So I think I could set it up.

bxabi:
I have 1-2 questions to clarify.
Are the files/modules which should be documented the eth_bridge.py, job.py, kvstore_abi.py and storage.py?
If that's right, it seems that sphinx imports the files, which means that it runs the code which is not in a function, and this causes the doc generation to fail.

When I put that code in a function, the doc generation works. So I could do this, but I suppose this effects other parts of the system which imports this code, because it would have to call an initialization method, which is currently executed just by importing.

rbval:
Hi @bxabi and thanks for the interest. The ones that need documenting are: `eth_bridge.py`, `job.py`, `storage.py`. `kvstore_abi.py` doesn't need to be documented.

rbval:
Would prefer not changing the existing logic with a wrapper function.

bxabi:
I don't know how this code is used, but it looks like a module which is imported by other code.
Isn't it bad practice in general to have code running at import (especially code connecting to network services)?

bxabi:
This is the section talking about the import: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#module-sphinx.ext.autodoc

[...]

PR #156, opened 2020-03-19
Docs/bxabi 72 generate docs

bxabi:
The changes required to automatically create documentation from doctests unsing sphinx.
To generate it, run 'make' in the sphinx-documentation folder.

### DIFF
--- a/bin/generate-docs
+++ b/bin/generate-docs
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# this script edits the sources allowing the autodoc to import them, runs the generation, then reverts the changes.
+
+format='html'
+if [ $# -eq 1 ] ; then
+  format=$1
+fi
+
+mv hmt_escrow/storage.py hmt_escrow/storage.py.backup
+mv hmt_escrow/eth_bridge.py hmt_escrow/eth_bridge.py.backup
+python sphinx-documentation/escape_code.py storage
+python sphinx-documentation/escape_code.py eth_bridge
+
+cd sphinx-documentation
+make $format
+cd ..
+
+mv hmt_escrow/storage.py.backup hmt_escrow/storage.py
+mv hmt_escrow/eth_bridge.py.backup hmt_escrow/eth_bridge.py
--- a/hmt_escrow/eth_bridge.py
+++ b/hmt_escrow/eth_bridge.py
@@ -83,8 +83,10 @@ def handle_transaction(txn_func, *args, **kwargs) -> AttributeDict:
 
     Args:
         txn_func: the transaction function to be handled.
-        *args: all the arguments the function takes.
-        **kwargs: the transaction data used to complete the transaction.
+
+        \*args: all the arguments the function takes.
+
+        \*\*kwargs: the transaction data used to complete the transaction.
 
     Returns:
         AttributeDict: returns the transaction receipt.
@@ -160,6 +162,7 @@ def get_escrow(escrow_addr: str) -> Contract:
     >>> job = Job(credentials=credentials, escrow_manifest=manifest)
 
     Deploying a new Job to the ethereum network succeeds.
+    
     >>> job.launch(rep_oracle_pub_key)
     True
     >>> type(get_escrow(job.job_contract.address))
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,9 +8,11 @@ Main repository we are starting the bounty program with is hmt-escrow. The proje
 
 ### Solidity
 The solidity project is based on truffle. The smart contracts can be found from the contracts/ folder.
-Python
+
+### Python
 The python part of the project is to act as a wrapper for the smart contracts since the main way we use the smart contracts is via python. You can find the python related part of the project inside the hmt_escrow/ folder.
-Development
+
+### Development
 We encourage using Docker as the main way to develop since it assembles all the solidity and python dependencies together neatly. Docker will be especially handy when developing python related modules. If you are only developing smart contract related features, a local ganache together with truffle should be sufficient.
 
 ### The First Issue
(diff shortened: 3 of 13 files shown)


rbval:
@bxabi any progress, blockers?

bxabi:
@rbval, yes. As I asked in the issue #72, there were code changes required, and I didn't get any recommendation how is the best way to do that, but I got encouraged to do a pull request, I decided to do that.
I have a new question now: is this project used from another projects, so this change would break them too, or it is enough to get the tests pass in this project alone?

rbval:
@bxabi yes, this is used by other projects. Let’s try not introducing breaking changes. Changing how this gets imported is one of them.

bxabi:
I added the generate_docs.sh script in the bin folder. This way the docs can be generated without code changes.

rbval:
Ok, will try to check this tomorrow.

bin/generate-docs.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
rbval: Check from other scripts how to name the first line. Also rename `generate-docs.sh` to `generate-docs`

hmt_escrow/eth_bridge.py
```diff
@@ -6,7 +6,7 @@
 from web3 import Web3, HTTPProvider, EthereumTesterProvider
 from web3.contract import Contract
 from web3.middleware import geth_poa_middleware
-from hmt_escrow.kvstore_abi import abi as kvstore_abi
```
rbval: All these should be put back since I remember we have upstream issues with relative imports

bin/generate-docs.sh
bxabi: I see other scripts having the exact same line, can you tell in more detail what you mean by naming the first line?

hmt_escrow/eth_bridge.py
bxabi: Using non-relative import here is an error and should not work, as it would use a different version of htm_escrow installed with pip or pipenv, not the one in the code.

As I don't have that module installed (since I'm working on it locally), it can't be imported, and that looks to be the correct behavior. So if that doesn't work on upstream that is probably a bug which should be fixed actually.
But maybe it does works, as the tests pass with the relative import.

bin/generate-docs.sh
rbval: They are `#!/bin/sh` / `#!/bin/bash` apparently. Only remove the `.sh` ending from the name.

hmt_escrow/eth_bridge.py
bxabi: I tried a couple things, and I found out that if I use the project folder as root instead of the python code folder, the non-relative import works also.

rbval:
Hi @bxabi , generating seems to work good, good job! 

Would be good to make the doc generation script docker ready as well so a person can just come to the project and run `bin/generate_doc`. Noticed now the issue what you talked earlier with not showing all code examples correctly:
image
Did you have an idea how to tackle that?

bxabi:
I think the layout is fixed if there is a new line between text and lines starting with ">>>", I changed that.  I will look at the docker readiness too.

bxabi:
Could you describe in more detail what the docker readiness means? (I'm not yet familiar with the structure of the project, and I'm not sure how and at which point should the doc-generation be connected to docker)

rbval:
Mainly that we could run a fresh version of docs each time we run `bin/ci`. That way any changes we do to doctests are always hand in hand with the latest docs. So just adding that one line inside `bin/ci` should do it.

bxabi:
Ok I added the ./bin/generate-docs to it.

bin/generate-docs
```diff
@@ -0,0 +1,20 @@
...
+format='html'
+if [ $# -eq 1 ] ; then
+  format=$1
+fi
+
+mv hmt_escrow/storage.py hmt_escrow/storage.py.backup
+mv hmt_escrow/eth_bridge.py hmt_escrow/eth_bridge.py.backup
+python sphinx-documentation/escape_code.py storage
+python sphinx-documentation/escape_code.py eth_bridge
+
```
gaieges, line 14: is this something we should do in code permanently?

bxabi: You mean if we should refractor the code so we don't need the escape_code script, or if the script still works if the code changes?

gaieges: Both, yes.  Seems odd to me that the code is tweaked temporarily to generate docs; does the generate function not like those lines?  Does the code break if its permanently changed in this way?

bxabi: As discussed above, I also think the script is a bad idea and the code should be fixed, as the files are not classes, and currently there is initialization code hanging around without being in a function. So it can't be imported without side-affect calls.

gaieges: That discussion being https://github.com/hCaptcha/hmt-escrow/issues/72#issuecomment-590786691 ?  

Gotcha; yeah I'd agree, probably scope for another pr

bxabi: Yes that sounds good. The places where they are used have to be then adjusted to instantiate or initialize these when needed. Where are they used?

rbval:
@bxabi this looks good to merge, I will do it myself at some point. Good job! If you are looking for the next issue, this one is a good one: https://github.com/hCaptcha/hmt-escrow/issues/160

gaieges:
Got a merge conflict here too, needs a rebase

bxabi:
I reorganized to commits so instead of all the changes and several merges it is just 5 smaller commits.

posix4e:
Anything blocking this merge?

posix4e:
Closing in favour of https://github.com/hCaptcha/hmt-escrow/pull/196

PR #158, opened 2020-03-23
Add multi-cred support

### LINKED ISSUE #155: Add support for multi-credentials
Give the launcher a possibility to specify multiple wallet credentials that can be used to launch a new contract. This lets the launcher launch multiple transactions without getting nasty "transaction replacement" issues.

Also, if a launch fails, randomly choose a wallet from the multi-cred dict.

1. Add optional multi-cred dict upon initialization
2. Randomly choose a new wallet if problems occur upon loading

rbval:
Closes #155 
Added initial version of the multi-cred support. Existing API doesn't break, adds only an enhancement with `multi_credentials`.

Design decisions:
1. Use a list of tuples for easy looping when raffling new credentials
2. Just walk the list instead of taking a random when raffling new credentials. Gave it some thought and this felt a lot simpler and more deterministic. Am I missing something?
3. Added new tests to test out that raffling indeed works and launches the contract as intended to the blockchain

### DIFF
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -911,7 +911,8 @@ class Job:
         calculated_addr = pub_key.to_checksum_address()
         return Web3.toChecksumAddress(addr) == calculated_addr
 
-    def _validate_credentials(self, multi_credentials: List[Tuple], **credentials) -> bool:
+    def _validate_credentials(self, multi_credentials: List[Tuple],
+                              **credentials) -> bool:
         """Validates whether the given ethereum private key maps to the address
         by calculating the checksum address from the private key and comparing that
         to the given address.
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -210,6 +210,7 @@ class Job:
     """
     def __init__(self,
                  credentials: Dict[str, str],
+                 multi_credentials: Dict[str, str] = None,
                  escrow_manifest: Manifest = None,
                  factory_addr: str = None,
                  escrow_addr: str = None):
@@ -285,10 +286,10 @@ class Job:
             ValueError: if the credentials are not valid.
 
         """
-        credentials_valid = self._validate_credentials(**credentials)
+        credentials_valid = self._validate_credentials(**credentials, **multi_credentials)
         if not credentials_valid:
             raise ValueError(
-                "Given private key doesn't match the ethereum address.")
+                "Given private keys don't match the ethereum addresses.")
 
         self.gas_payer = Web3.toChecksumAddress(credentials["gas_payer"])
         self.gas_payer_priv = credentials["gas_payer_priv"]
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -1128,6 +1128,8 @@ class Job:
             }
             try:
                 handle_transaction(txn_func, *[], **txn_info)
+                self.gas_payer = gas_payer
+                self.gas_payer_priv = gas_payer_priv
                 escrow_created = True
                 break
             except:
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -914,6 +914,31 @@ class Job:
 
     def _validate_multi_credentials(
             self, multi_credentials: List[Tuple]) -> List[Tuple[Any, Any]]:
+        """Validates whether the given ethereum private key maps to the address
+        by calculating the checksum address from the private key and comparing that
+        to the given address.
+
+        >>> credentials = {
+        ...     "gas_payer": "0x1413862C2B7054CDbfdc181B83962CB0FC11fD92",
+        ... 	"gas_payer_priv": "28e516f1e2f99e96a48a23cea1f94ee5f073403a1c68e818263f0eb898f1c8e5"
+        ... }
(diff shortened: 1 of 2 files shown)


hmt_escrow/job.py
```diff
@@ -212,7 +212,8 @@ def __init__(self,
                  credentials: Dict[str, str],
                  escrow_manifest: Manifest = None,
                  factory_addr: str = None,
-                 escrow_addr: str = None):
+                 escrow_addr: str = None,
+                 multi_credentials: list = []):
```
gaieges: Might be good to add a `list[Tuple]` to the type here

hmt_escrow/job.py
```diff
@@ -911,13 +940,24 @@ def _validate_credentials(self, **credentials) -> bool:
...
-        priv_key_bytes = decode_hex(priv_key)
-        pub_key = keys.PrivateKey(priv_key_bytes).public_key
-        calculated_addr = pub_key.to_checksum_address()
-        return Web3.toChecksumAddress(addr) == calculated_addr
+        addr_valid = self._eth_addr_valid(gas_payer_addr, gas_payer_priv)
+
+        if not multi_credentials:
+            return addr_valid
+
+        for eth_addr, priv_key in multi_credentials:
```
gaieges: Do we care if they're ALL valid?  Really just care about the one we're going to use, no?

rbval: I think it's good practice to let the person know in the setup step if they've typo'd the private key for example.

hmt_escrow/job.py
rbval: Fixed

hmt_escrow/job.py
gaieges: I agree, always good to give more info .. but will 1 bad key out of 10 block a deploy?

rbval: Yea it will block. I'd still say it's still good since it'll log it. 

Also, it's not as if we constantly change keys so it should be really few times a year thing.

gaieges: Why not just log the bad ones and take out of the pool of available ones to deploy with; that way we learn about the problem, and don't break any deploys at the same time

rbval: Yeah, maybe credentials validation makes more sense with the "main credentials". For the other ones we could fail by logging and raffle new ones.

Only thing I'm thinking is that is it more probable that we fuck up only 1 key out of 10 or it's something more systemic where we fuck up all the keys?

Just a bit weird to me to validate main credentials but not validating the rest of them.

gaieges: > Only thing I'm thinking is that is it more probable that we fuck up only 1 key out of 10 or it's something more systemic where we fuck up all the keys?

I would think the former, like via a human improper copy / paste.

I'd validate em all, if we could!  Do we need to differentiate main vs secondary?

rbval: "I'd validate em all, if we could!" what do you mean? This does exactly that now.

gaieges: Oh good, thats what I was thinking.  Was thrown off by:

> Just a bit weird to me to validate main credentials but not validating the rest of them.

rbval: Main reason I want to distinguish "credentials" from "multi-credentials" is to not introduce breaking changes. That way we can use it with and without multi-credentials.

gaieges: Oh I gotcha, that makes sense.  So validate all, if main fails, abort; if remainder fails, proceed but warn.. right?

rbval: So when when we do this:
`job = Job(credentials, manifest, multi_credentials=multi_credentials)`
we will currently fail that if `multi_credentials` contains wrong eth_addr:priv_key pairs (IMO this is not an issue since we should anyways check carefully that we use correct eth_addr:priv_key pairs)

After we know all keys are valid, we first try to do the escrow creation with `credentials` (main). If that fails we start going through `multi_credentials` and raffle until we succeed. So we will keep new creds until we succeed.

gaieges: Yeah I'm still just saying that if `multi_credentials` has at least 1 working credential, we should use it and ignore the ones that aren't valid; and throw a warning that theres a bad credential in multi_creds

rbval: So should `_validate_credentials` just check there's at least one valid credential in `multi_credentials` and return true?

gaieges: Probably just return a list of valid ones for use, warn on the bad ones.  Raise exception / return false if none?

rbval: Lol, actually we don't even validate `multi_credentials` in `_validate_credentials` so wrong alert! Works exactly as you described.

rbval: We will anyways catch those in `_create_escrow` phase. The only edge case would be if ALL the creds were falsy. I think we are nitpicking quite a lot here.

gaieges: Oh lol. yeah I'm not really too concerned with the code itself more the behavior of a single bad credential in the list shouldnt fail the entire deploy.  :)

rbval:
We have the doctests that will show how to use multi-cred. Separately, we have @bxabi doing general docs in another PR: https://github.com/hCaptcha/hmt-escrow/pull/156

PR #159, opened 2020-03-26
Add trusted handlers

### LINKED ISSUE #149: Add a collection of "trusted handlers" that can perform actions behalf of the original contract deployer.  Support canceling of jobs
Thinking something like providing a list of possible gas_payers with pub/priv keys so that we can eliminate problems that may occur when deploying multiple jobs at once.

Q: do we do this with requester keys as well?

cc: @rbval @posix4e

rbval:
Closes #149 

Lets us specify separate cancelers on launch time, so called "trusted handlers". These will be able to abort jobs on behalf of the launcher.

### DIFF
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -601,6 +601,26 @@ class Job:
         >>> job.status()
         <Status.Paid: 4>
 
+
+        Trusted handler should be able to abort an existing contract
+        >>> trusted_handler = "0x6b7E3C31F34cF38d1DFC1D9A8A59482028395809"
+        >>> job = Job(credentials, manifest)
+        >>> job.launch(rep_oracle_pub_key)
+        True
+        >>> job.setup()
+        True
+        >>> job.add_trusted_handlers([trusted_handler])
+        True
+
+        >>> handler_credentials = {
+        ... 	"gas_payer": "0x6b7E3C31F34cF38d1DFC1D9A8A59482028395809",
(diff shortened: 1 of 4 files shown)


posix4e:
@rbval wanna add a test?

hmt_escrow/job.py
```diff
@@ -517,39 +572,55 @@ def abort(self, gas: int = GAS_LIMIT) -> bool:
...
-        >>> payouts = [("0x852023fbb19050B8291a335E5A83Ac9701E7B4E6", Decimal('80.0'))]
+        >>> payouts = [("0x852023fbb19050B8291a335E5A83Ac9701E7B4E6", Decimal('100.0'))]
         >>> job.bulk_payout(payouts, {'results': 0}, rep_oracle_pub_key)
         True
         >>> job.abort()
         False
         >>> job.status()
         <Status.Paid: 4>
 
+
```
rbval: @posix4e it's here.

rbval:
@gaieges we have the launcher set as default, otherwise wouldn't put any defaults on smart contract level nor on the python API level. It's one line more in loader :)

gaieges:
Yeah I understand that we'd do it in loader, I'm wondering what 3rd party users would do if they're trying to run on our system? Probably not something to worry about now just wondering your thoughts

rbval:
They'd have to specify those via the python API yeah, just like we do.

rbval:
@posix4e added a JS smart contract level test as well.

PR #163, opened 2020-04-08
Feat 162 raffle payout and results

### LINKED ISSUE #162: Add support for credential raffling to bulk payout and store results

rbval:
Closes #162 
First part of making sure multiple wallets can store results and do bulk payments.
Also set rec and rep oracle as "trusted handlers" in setup time by default.

### DIFF
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -750,6 +750,18 @@ class Job:
         >>> rep_oracle_priv_key = b"28e516f1e2f99e96a48a23cea1f94ee5f073403a1c68e818263f0eb898f1c8e5"
         >>> job.intermediate_results(rep_oracle_priv_key)
         {'results': True}
+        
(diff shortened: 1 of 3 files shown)


hmt_escrow/job.py
```diff
@@ -1283,4 +1360,4 @@ def _create_escrow(self, gas: int = GAS_LIMIT) -> bool:
 if __name__ == "__main__":
     import doctest
     from test_manifest import manifest
-    doctest.testmod(raise_on_error=True)
```
gaieges, line 1286: this is what failed CI before, IIRC.  If this is not set it may fail and not `exit 1` thus failing a build

rbval: Ah yes, always locally disabling to see track traces.

gaieges: Can you also throw a comment in there to remind ppl?  I randomly remembered this and still not 100% sure on it

PR #179, opened 2020-05-05
Add raffling to setup, make job loading faster

### LINKED ISSUE #161: Add trusted handlers by default when creating an escrow

### LINKED ISSUE #172: Protect against flakiness between launch and setup: contract did not get funded
If there's an issue between launch and setup, we should raffle new credentials, we see it as `contract did not get funded` in logs. Thus we should support adding trusted handlers launch time to be more resilient.

rbval:
Closes #172 
Closes #161

1. Adds raffling on setup
2. By default adds our multi creds to the contract when we launch the contract

Benefits:
1. We should not see `contract did not get funded` anymore as we used to.
2. Make job loading 15 seconds faster (per job load)

### DIFF
--- a/contracts/Escrow.sol
+++ b/contracts/Escrow.sol
@@ -40,7 +40,12 @@ contract Escrow {
 
     mapping(address => TrustedHandler) public trustedHandlers;
 
-    constructor(address _eip20, address _canceler, uint256 _expiration) public {
+    constructor(
+        address _eip20,
+        address _canceler,
+        uint256 _expiration,
+        address[] _handlers
+    ) public {
         eip20 = _eip20;
         status = EscrowStatuses.Launched;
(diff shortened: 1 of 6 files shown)


contracts/Escrow.sol
```diff
@@ -40,14 +40,20 @@ contract Escrow {
 
     mapping(address => TrustedHandler) public trustedHandlers;
 
-    constructor(address _eip20, address _canceler, uint256 _expiration) public {
+    constructor(
+        address _eip20,
+        address _canceler,
+        uint256 _expiration,
+        address[] _handlers
```
gaieges, line 47: This means we'll need to update all the contracts, correct?

rbval: yes

PR #181, opened 2020-05-06
Add an option for s3 storage

### LINKED ISSUE #180: Move python API to use s3 by default instead of IPFS
We have been experiencing broad issues with the reliability of IPFS. Move our `upload` and `download` functions to s3 and use the `manifest_hash` instead of `manifest_url`.

rbval:
Closes #180 

Add an option for S3 storage for download and upload. Doesn't introduce an API change since we default to s3.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,7 @@ services:
       - ESCROW_AWS_ACCESS_KEY_ID=minio
       - ESCROW_AWS_SECRET_ACCESS_KEY=minio123
       - ESCROW_ENDPOINT_URL=http://minio:9000
+      - ESCROW_AWS_DEFAULT_REGION=us-west-2
       - DEBUG=true
     command: /work/bin/wait_then_run
     build: .
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,12 +18,13 @@ services:
     command: /work/bin/wait_then_run
     build: .
     image: hmtcentral.azurecr.io/contracts:latest
+    depends_on:
+      - minio
     links:
       - ganache
       - ipfs
       - hmt
       - kvstore
-      - minio
     volumes:
       - hmt_token_addr:/deployed-hmtoken
       - eth_kvstore_addr:/deployed-kvstore
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,6 @@ services:
       - ESCROW_ENDPOINT_URL=http://minio:9000
       - ESCROW_AWS_DEFAULT_REGION=us-west-2
       - DEBUG=true
-      - LOCAL=true
     command: /work/bin/wait_then_run
     build: .
     image: hmtcentral.azurecr.io/contracts:latest
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,11 @@ services:
       - GAS_LIMIT=6700000
       - IPFS_HOST=ipfs
       - IPFS_PORT=5001
-      - IPNS_PATH=/ipns/
+      - ESCROW_BUCKETNAME=escrow-results
+      - ESCROW_AWS_ACCESS_KEY_ID=minio
+      - ESCROW_AWS_SECRET_ACCESS_KEY=minio123
+      - ESCROW_ENDPOINT_URL=http://minio:9000
+      - DEBUG=true
     command: /work/bin/wait_then_run
     build: .
     image: hmtcentral.azurecr.io/contracts:latest
@@ -19,6 +23,7 @@ services:
       - ipfs
       - hmt
       - kvstore
+      - minio
     volumes:
       - hmt_token_addr:/deployed-hmtoken
       - eth_kvstore_addr:/deployed-kvstore
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,6 +28,7 @@ services:
     volumes:
       - hmt_token_addr:/deployed-hmtoken
       - eth_kvstore_addr:/deployed-kvstore
+      - ./minio:/data
       - .:/work
 
   contracts:
@@ -82,11 +83,9 @@ services:
     environment:
       MINIO_ACCESS_KEY: minio
       MINIO_SECRET_KEY: minio123
-    entrypoint: sh -c 'cp -a /template/. /data && minio server /data/'
+    entrypoint: sh -c 'minio server /data/'
     volumes:
-      - ./minio:/template
-    tmpfs:
-      - /data
+      - ./minio:/data
 
 volumes:
   hmt_token_addr:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,7 @@ services:
       - ESCROW_ENDPOINT_URL=http://minio:9000
       - ESCROW_AWS_DEFAULT_REGION=us-west-2
       - DEBUG=true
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -10,10 +10,16 @@ services:
       - GAS_LIMIT=6700000
       - IPFS_HOST=ipfs
       - IPFS_PORT=5001
-      - IPNS_PATH=/ipns/
+      - ESCROW_BUCKETNAME=escrow-results
+      - ESCROW_AWS_ACCESS_KEY_ID=minio
+      - ESCROW_AWS_SECRET_ACCESS_KEY=minio123
+      - ESCROW_ENDPOINT_URL=http://minio:9000
```
posix4e, line 16: how do we control what aws region to hit?

rbval: good point, adding

hmt_escrow/storage.py
```diff
@@ -10,14 +10,33 @@
...
     b'9da0d3721774843193737244a0f3355191f66ff7321e83eae83f7f746eb34350')
 
+logging.getLogger("boto").setLevel(logging.INFO)
+logging.getLogger("botocore").setLevel(logging.INFO)
+logging.getLogger("boto3").setLevel(logging.INFO)
+
 LOG = logging.getLogger("hmt_escrow.storage")
+DEBUG = "true" in os.getenv("DEBUG", "false").lower()
+LOG.setLevel(logging.DEBUG if DEBUG else logging.INFO)
+LOCAL = os.getenv("LOCAL", False)
```
gaieges, line 27: might wanna comment on what this means .. I see the comment below but its a bit nebulous up here.

hmt_escrow/storage.py
```diff
@@ -10,14 +10,33 @@
...
+
 IPFS_HOST = os.getenv("IPFS_HOST", "localhost")
 IPFS_PORT = int(os.getenv("IPFS_PORT", 5001))
 
+ESCROW_BUCKETNAME = os.getenv("ESCROW_BUCKETNAME", "escrow-results")
+ESCROW_AWS_ACCESS_KEY_ID = os.getenv("ESCROW_AWS_ACCESS_KEY_ID", "minio")
+ESCROW_AWS_SECRET_ACCESS_KEY = os.getenv("ESCROW_AWS_SECRET_ACCESS_KEY",
+                                         "minio123")
+ESCROW_AWS_REGION = os.getenv("ESCROW_AWS_REGION", "us-west-2")
+ESCROW_AWS_KMS = os.getenv("ESCROW_AWS_KMS", "abcd1234-efgh5678")
```
gaieges, line 37: did we want kms here?

hmt_escrow/storage.py
```diff
@@ -28,10 +47,39 @@ def _connect(host: str, port: int) -> Client:
...
+            "ServerSideEncryption": "aws:kms",
+            "SSEKMSKeyId": ESCROW_AWS_KMS,
+            "Bucket": ESCROW_BUCKETNAME,
+            "Key": s3_hash,
+            "Body": encrypted_msg
+        }
 
-def download(key: str, private_key: bytes) -> Dict:
+
+def download(key: str, private_key: bytes, s3: bool = True) -> Dict:
```
gaieges, line 82: I'm curious why we need the s3 arg here.. can't we support both by just looking at the key itself?  (ie if `s3://...` then use s3, if not, default to ipfs?). Supports both and allows for backward compatibility while we roll this out, and doesn't change the API

hmt_escrow/storage.py
rbval: removed

hmt_escrow/storage.py
rbval: The key here would be the hash we store in the contract and inside the bucket with the following encryption:
`hash_ = hashlib.sha1(manifest_.encode('utf-8')).hexdigest()` which will act as an index inside the bucket. 

The hash looks something like this: `3399620f41cd85c362e4b59a18c15e23cf746c66`

Currently this doesn't change the API since we would be defaulting to s3, it will only change if we ever want to switch back to IPFS.

hmt_escrow/storage.py
rbval: removed the whole variable since it's not needed anymore

hmt_escrow/storage.py
gaieges: It does change the api though, since you're defaulting to something that is different from what it was before.

rbval: I guess an `s3` prefix could be added but that then again will be exposed to the public via the contract, anyone could see it. 

I see the point though of having a prefix that could be used to automatically choose between ipfs and s3 when downloading. When uploading we would however have to define the strategy (s3 / ipfs).

rbval: The outside API doesn't change, just the end destination of the content (ipfs -> s3). We would not need to change anything in the code.

gaieges: The interface doesn't change, but the behavior does.  its a breaking change since I can't just upgrade this in place in my existing code, I need to go specify `s3=False` for the same behavior

> When uploading we would however have to define the strategy (s3 / ipfs).

That's true, I would expect the default to be ipfs and if s3=true then do the s3 upload.

Could always just increase the minor version as well; my bigger concern is just making it backward compatible so we dont have cases where as soon as we upgrade we can't do anything with the old jobs that are still ipfs based

rbval: What if we added two functions: `s3_upload` `s3_download`? Would be probably the clearest without more advanced if logic while guaranteeing that we maintain backward compatibility?

rbval: The other option would be to default to IPFS as you mentioned.

gaieges: With the two functions you have a similar problem .. you have to have the calling code differentiate between "is it s3 or is it ipfs" which would work just need to do more on the calling side.

rbval: Changes made:
1. `upload` constructs an `s3` prefix to the `manifest_url` if s3 has been chosen. Defaults to ipfs
2. `download` automatically determines whether to choose ipfs or s3 based on the prefix

hmt_escrow/storage.py
```diff
@@ -56,19 +88,35 @@ def download(key: str, private_key: bytes) -> Dict:
...
-        raise e
-    return json.loads(msg)
-
-
-def upload(msg: Dict, public_key: bytes) -> Tuple[str, str]:
+    if key.startswith("s3"):
+        try:
+            LOG.debug("Downloading s3 key: {}".format(key))
+            BOTO3_CLIENT = _connect_s3()
+            response = BOTO3_CLIENT.get_object(Bucket=ESCROW_BUCKETNAME,
```
gaieges, line 95: might be handy to break these into two separate try/catches, one that does the fetching, one that does the decrypting, so the log message is targeted.  not blocking though

hmt_escrow/storage.py
```diff
@@ -165,4 +238,4 @@ def _encrypt(public_key: bytes, msg: str) -> bytes:
     import doctest
     from test_manifest import manifest
     from job import Job
-    doctest.testmod(raise_on_error=True)
```
gaieges, line 168: dont forget about this. also there must be another way to do this on the cli so we dont have to change code every time

PR #187, opened 2020-05-12
Fix 186 noisy raffling

### LINKED ISSUE #186: Make raffling less noisy
We give an error log for every failed raffle. Only give an error if all raffling attempts fail, not just one.

rbval:
Closes #186

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -75,9 +75,7 @@ services:
       - 5001:5001
   
   minio:
-    image: minio/minio
-    ports:
-      - "9000:9000"
+    image: minio/minio:9423-d242680
     expose:
       - "9000"
     environment:
@@ -86,6 +84,8 @@ services:
     entrypoint: sh -c 'cp -a /template/* /data && minio server /data/'
     volumes:
       - ./minio:/template
+    tmpfs:
+      - /data
 
 volumes:
   hmt_token_addr:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -76,7 +76,7 @@ services:
       - 5001:5001
   
   minio:
-    image: minio/minio:9423-d242680
+    image: minio/minio:RELEASE.2020-04-28T23-56-56Z
     ports:
       - "9000:9000"
     expose:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -84,9 +84,9 @@ services:
     environment:
       MINIO_ACCESS_KEY: minio
       MINIO_SECRET_KEY: minio123
-    entrypoint: sh -c 'minio server /data/'
+    entrypoint: sh -c 'cp -a /template/* /data && minio server /data/'
     volumes:
-      - ./minio:/data
+      - ./minio:/template
 
 volumes:
   hmt_token_addr:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -76,7 +76,7 @@ services:
       - 5001:5001
   
   minio:
-    image: minio/minio
+    image: minio/minio:9423-d242680
```
gaieges, line 79: do we need to lock this?

hmt_escrow/job.py
```diff
@@ -1432,9 +1436,10 @@ def _create_escrow(self,
                 escrow_created = True
                 break
             except Exception as e:
-                LOG.error(
-                    f"Contract creation failed with {gas_payer} and {gas_payer_priv} due to {e}. Raffling new ones..."
-                )
+                if gas_payer == self.multi_credentials[-1][0]:
+                    LOG.error(
```
gaieges, line 1440: can you make this a `LOG.exception`?

hmt_escrow/job.py
```diff
@@ -1432,9 +1436,10 @@ def _create_escrow(self,
                 escrow_created = True
                 break
             except Exception as e:
-                LOG.error(
-                    f"Contract creation failed with {gas_payer} and {gas_payer_priv} due to {e}. Raffling new ones..."
-                )
+                if gas_payer == self.multi_credentials[-1][0]:
```
gaieges, line 1439: perhaps here you should just use an iterator in the loop?  I fear this check might be a bit fragile

docker-compose.yml
rbval: yeah, didn't work with latest, used from recoracle

hmt_escrow/job.py
rbval: sure

hmt_escrow/job.py
rbval: yea, good idea

docker-compose.yml
gaieges: Ahh, is there a better pinned version that we can use? Like v0.3 or something?  Looking through their I just see stuff like `minio/minio:RELEASE.2019-02-06T21-16-36Z` which is basically the same as what you're doing there (build specific vs minor version specific)

hmt_escrow/job.py
gaieges: Hmm one other thing I was thinking here was - is there any action to take aside from the logging here? Ie do we cancel the contract? Or just not create it at all and throw exception?

rbval: We'll retry it the next iteration if this time it fails as the status of the contract won't change

docker-compose.yml
rbval: I just took this version from recoracle, I guess we could use something else

hmt_escrow/job.py
rbval: Cancelling is probably too aggressive.

hmt_escrow/job.py
rbval: going to use already defined boolean fields to perform this, more elegant.

docker-compose.yml
rbval: Now it's pinned to the latest release of minio that works.

gaieges: Cool yeah no worries, probably about the same as what you had, they generally have better floating version tags (not in minio's case it looks like)

PR #189, opened 2020-05-14
Add raffling to the remaining transaction functions

### LINKED ISSUE #188: Add raffling to the rest of the job.py functions
Next thing we'll need raffling to is `abort`. However it would be great at this point to just add it to all functions directly as the logic is not that big of an addition.

Add support for raffling to the rest of the functions inside `job.py`

rbval:
Closes #188 

1. Use an internal function to keep the code DRY
2. Add raffling to remaining functions that perform a transaction


rbval:
@gaieges yea they catch these quite well, will do a dry-run locally for just in case before merging though

PR #204, opened 2020-07-10
Reduce Escrow gas usage

### LINKED ISSUE #197: Reduce Escrow gas usage
Currently we are using too much gas. With our move from IPFS to s3, we shouldn't update the contract nearly as much.

TODO
- [x] Remove intermediate results hash and url
- [x] Only update the URL on start, and potentially on finish
- [x] Only update the hash when the contract is complete

Now store intermediate results should only use gas on the first call. This should greatly reduce gas use

uivlis:
Fixes #197.

### DIFF
--- a/contracts/Escrow.sol
+++ b/contracts/Escrow.sol
@@ -1,4 +1,5 @@
 pragma solidity 0.6.2;
+
 import "./HMTokenInterface.sol";
 import "./SafeMath.sol";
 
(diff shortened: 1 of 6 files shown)


posix4e:
Ok so just to be clear, the first time they store a chunk it updates the URL? Then we update the data under the hood. And as a final step we set the final hash value?

uivlis:
Yes!

posix4e:
> Yes.

Cool I'll try it out

PR #205, opened 2020-07-10
Measure gas

### LINKED ISSUE #201: Measure, simulate and understand gas usage
Be nice to have a test, which measures gas usage on different changes to simulate costs

uivlis:
Fixes #201.

### DIFF
--- a/contracts/Escrow.sol
+++ b/contracts/Escrow.sol
@@ -1,4 +1,5 @@
 pragma solidity 0.6.2;
+
 import "./HMTokenInterface.sol";
 import "./SafeMath.sol";
 
@@ -21,9 +22,6 @@ contract Escrow {
 
     string private manifestUrl;
     string private manifestHash;
-
-    string private intermediateResultsUrl;
-    string private intermediateResultsHash;
 
     string private finalResultsUrl;
     string private finalResultsHash;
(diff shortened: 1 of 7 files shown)


posix4e:
This a good start. How do the gas costs display? What does it look like?

uivlis:
During testing, it logs to console something like this:
```
...
      ✓ pays each recipient their corresponding amount (427ms)
Setup costs: 164716 wei.
BulkPayOut costs: 202482 wei.
Complete costs: 29608 wei.
...
```

posix4e:
Would it be reasonable to set some max cumulative gas as a part of testing?

uivlis:
You mean the tests to fail if max total gas reached? That would mean to define a max total gas for approx. 30 diferent tests.

posix4e:
> You mean the tests to fail if max total gas reached? That would mean to define a max total gas for approx. 30 diferent tests.

Or maybe just one or two of the tests that we wanna assure. Just something that's less than manual

uivlis:
I don't even know what standard to use when setting those limits... Do you have some limits in mind?

posix4e:
> I don't even know what standard to use when setting those limits... Do you have some limits in mind?

Basic sanity checks. Pick a couple of things that would be bad to say increase by 5x.

uivlis:
Sanity checks, checked.

PR #215, opened 2020-07-30
Update web3 and py-evm using trinity

uivlis:
Updated hmt-escrow to use the latest versions of web3, py-evm and p2p (commonly referred to recently as 'trinity') needed for Python3.8. Revert to using pipenv.

### DIFF
--- a/bin/ci
+++ b/bin/ci
@@ -4,6 +4,5 @@ set -exu
 docker-compose build
 docker-compose run --no-deps -e CI=true job './bin/lint'
 docker-compose run -w /work job
-docker-compose run -w /work contracts
 ./bin/stop
 ./bin/generate-docs
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ import setuptools
 
 setuptools.setup(
     name="hmt-escrow",
-    version="0.8.2",
+    version="0.8.3",
     author="HUMAN Protocol",
     description="A python library to launch escrow contracts to the HUMAN network.",
     url="https://github.com/hCaptcha/hmt-escrow",
--- a/setup.py
+++ b/setup.py
@@ -14,5 +14,5 @@ setuptools.setup(
         "Programming Language :: Python",
     ],
     packages=setuptools.find_packages() + ["contracts", "migrations"],
-    install_requires=["py-solc-x", "web3", "hmt-basemodels", "boto3"],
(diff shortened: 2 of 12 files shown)


setup.py
```diff
@@ -14,5 +14,5 @@
         "Programming Language :: Python",
     ],
     packages=setuptools.find_packages() + ["contracts", "migrations"],
-    install_requires=["py-solc-x", "web3", "hmt-basemodels", "boto3"],
+    install_requires=["py-solc-x", "trinity", "hmt-basemodels", "boto3", "sphinx"],
```
gaieges, line 17: I think we also need the specific web3 version here btw.

Also you will need to bump the hmt-escrow version in this file

uivlis: So, web3 is contained in trinity now, but yes, hmt-escrow version is must. Will add that now.

gaieges: Ah I see, but not specifying a trinity version will always get you the latest in downstream packages, don't think we want that.

bin/ci
```diff
@@ -4,6 +4,5 @@ set -exu
 docker-compose build
 docker-compose run --no-deps -e CI=true job './bin/lint'
 docker-compose run -w /work job
-docker-compose run -w /work contracts
```
gaieges, line 7: Ahh didn't see this before, I think this is what @posix4e is referring to

uivlis:
@posix4e Nope, we're running the same tests...

bin/ci
uivlis: Hmm, I don't think so @gaieges. Anyway, I am running all tests, this was just a leftover from the previous hmt configuration.

uivlis:
Great, finally Travis builds successfully.

setup.py
uivlis: Actually, we kinda want it, since trinity has been fixed for some focal python bug only a few days ago, and the latest release does not contain that patch, yet.

gaieges: The problem it introduces is that it will allow future installs to get free major upgrades, which may have different API's and thus break code that depends on this. Specifying at least a major version is important.

uivlis: Will it do if I solemnly swear to specify a major version as soon as the commit from yesterday in master of trinity will get released?

gaieges: Yeah definitely, I would expect their release process to have labeled "latest" and the version at the same time but maybe that didnt happen.

PR #225, opened 2020-09-03
Adjust contracts as per latest audit report

uivlis:


e271828-:
Looks OK at first glance, but in the future please do not make such large commits. 

1. Each change from an audit report should be in its own commit, and the commit message should reference the report section.
2. Unrelated refactoring (e.g. renaming of variables) should be in a separate commit.

In this case, the audit report had only a few minor non-security-relevant comments that requested only a few lines of change. Identifying the matching changes in a several hundred line diff and confirming correctness of unrelated changes unnecessarily slows down review.

uivlis:
Hmm, I had no unrelated renamings, all of the changes were suggested in the report. But yes, you're right, I didn't think how difficult it would be for an evaluator to check all of the changes at once... For me, that may be a more generalizable takeaway, i.e., to break down commits into smaller chunks. Thanks for the review, let me know when you want this merged.

posix4e (approved):

truffle.js
```diff
@@ -14,20 +14,20 @@ module.exports = {
...
     },
     ropsten: {
-      provider: () => new HDWalletProvider(MNEMONIC, `https://ropsten.infura.io/${INFURA_TOKEN}`),
+      provider: () => new HDWalletProvider(MNEMONIC, `https://ropsten.infura.io/v3/${INFURA_TOKEN}`),
       network_id: '3',
       gas: 4700000,
     },
     rinkeby: {
-      provider: () => new HDWalletProvider(MNEMONIC, `https://rinkeby.infura.io/${INFURA_TOKEN}`),
+      provider: () => new HDWalletProvider(MNEMONIC, `https://rinkeby.infura.io/v3/${INFURA_TOKEN}`),
```
gaieges, line 30: We provide this token in other repos correct? I think best to leave this as it was before so we don't go needing to change it elsewhere

deployed-hmtoken/hmt.address.json
```diff
@@ -0,0 +1,3 @@
+{
+  "address": "0x4C18A2E51edC5043e9c4B6b0757990A4Ac13797f"
+}
```
gaieges, line 3: I dont think we want this checked in do we? this is an artifact of the token deploy

contracts/SafeMath.sol
```diff
@@ -1,66 +1,159 @@
...
+     * division by zero. The result is rounded towards zero.
+     *
+     * Counterpart to Solidity's `/` operator. Note: this function uses a
+     * `revert` opcode (which leaves remaining gas untouched) while Solidity
+     * uses an invalid opcode to revert (consuming all remaining gas).
+     *
+     * Requirements:
+     *
+     * - The divisor cannot be zero.
+     */
```
gaieges, line 118: Nice addition of comments in this file

contracts/HMToken.sol
```diff
@@ -153,7 +151,7 @@ contract HMToken is HMTokenInterface {
         if (_to == address(this)) return false; // Preclude sending tokens to the contract.
         if (balances[msg.sender] < _value) return false;
 
-        balances[msg.sender] = balances[msg.sender].sub(_value);
+        balances[msg.sender] = balances[msg.sender] - _value;
```
gaieges, line 154: Is the `-` better than the safemath version?

gaieges (commented):
Some thoughts, not super knowledgable in solidity though so I'd take it with a grain of salt

truffle.js
uivlis: That is not an option. Infura has upgraded their API to v3, and without that addition, the link won't work. The others, if not changed, won't work either when deploying.

contracts/SafeMath.sol
uivlis: Thanks, but they're not mine, I copy-pasted the SafeMath.sol file from openzeppelin's contracts, just like the other version (which was an older one, though). This is needed since solcx cannot compile files from different directories.

contracts/HMToken.sol
uivlis: It's better from a clarification perspective, meaning: you check for underflows in the above `if`, so it's useless to use that function from the SafeMath library (I think it's even a bit more gas-costly? not sure, though). Either I dropped the `if` (should I do that?), either I obeyed the recommendation from the certiK report (drop the `sub`).

deployed-hmtoken/hmt.address.json
uivlis: Okay, removing this.

contracts/HMToken.sol
gaieges: @posix4e might have an opinion but seems reasonable to me

posix4e (approved):

truffle.js
gaieges: Ahh, and the token remains the same there?  That sounds good then

PR #230, opened 2020-10-07
Added JS Package

excerebrose:
Adds a Javascript package to interact directly with blockchain & upload jobs
@gaieges @uivlis


hmt_escrow_js/server.js
```diff
@@ -0,0 +1,60 @@
+require('dotenv').config()
```
gaieges, line 1: We probably don't need the server file in here do we?

hmt_escrow_js/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "hmt-escrow-form-js",
+  "version": "1.0.0",
+  "description": "JS File for HMT-ESCROW-FORM",
+  "main": "server.js",
```
gaieges, line 5: The main would be the main entrypoint for the package, so probably an `index.js` file or something that exports `{ Job, EthInterface }` right?

It'd be good to see in your form something like:

```
import { Job } from 'hmt-escrow';
Job(..)
```

hmt_escrow_js/manifest-schema-pydantic.js
```diff
@@ -0,0 +1,602 @@
+module.exports = {
```
gaieges, line 1: Nice! Can you quickly throw in a comment in the readme on how you generated this?

hmt_escrow_js/Readme.md
```diff
@@ -0,0 +1,38 @@
+# HMT-Escrow Javascript Module
```
gaieges, line 1: ❤️

gaieges (commented):
Nice this is awesome! There's one little readme comment in here I'd block on, and probably the index.js thing too

I think we'll probably want to do some next steps in a separate bounty:
- Tests
- Create a npm package from this
- Break out to another repo (maybe)
- Add another consumer of this for validation that it works in multiple places.

For now awesome.  Want @posix4e and @uivlis  to take a look at this too from a blockchain perspective

hmt_escrow_js/package.json
excerebrose: fixed

hmt_escrow_js/manifest-schema-pydantic.js
excerebrose: done

hmt_escrow_js/server.js
excerebrose: removed, you're right

excerebrose:
@excerebrose changes done as requested

hmt_escrow_js/ETHInterface.js
```diff
@@ -0,0 +1,146 @@
+require('dotenv').config()
+
+const EscrowABI= require('./ABIs/Escrow')
+const EscrowFactoryABI = require('./ABIs/EscrowFactory')
+const HMTTokenInterfaceABI = require('./ABIs/HMTTokenInterface')
+
```
uivlis, line 6: You should fetch them from the artifacts, the ABI is subject to change, and on top of that, it shouldn't be uploaded. I think if you run `truffle compile` on the contract folder, the builded artifacts will be there, you should fetch them from there.

hmt_escrow_js/ETHInterface.js
```diff
@@ -0,0 +1,146 @@
...
+const HDWalletProvider = require("truffle-hdwallet-provider")
+const web3 = require('web3')
+const isUrl = require('is-url')
+
+let HMTOKEN_ADDR = process.env.HMTOKEN_ADDR || "0x4C18A2E51edC5043e9c4B6b0757990A4Ac13797f"
```
uivlis, line 11: Just a kinky note, but this is not our current HMTOKEN_ADDR, which is `0x4dCf5ac4509888714dd43A5cCc46d7ab389D9c23`.

hmt_escrow_js/ETHInterface.js
```diff
@@ -0,0 +1,146 @@
...
+
+class ETHInterface {
+  constructor(MNEMONIC, INFURA_KEY, NETWORK) {
+    // Setups Required
+    if (!MNEMONIC || !INFURA_KEY || !NETWORK ) {
+      console.error("Please set a mnemonic, infura key,and network")
+      return
+    }
+    const provider = new HDWalletProvider(MNEMONIC, `https://${NETWORK}.infura.io/v3/${INFURA_KEY}`)
```
uivlis, line 20: You should also allow for the option to connect to Infura via Websocket, I recently had an issue with that, meaning I couldn't query the contract for events, since those RPC calls are available only for Websocket, not HTTPS. But this is really just a note for the future, no need to do that now.

hmt_escrow_js/Job.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  _process_manifest() {
+    this.amount = Math.round(this.serialized_manifest.task_bid_price * this.serialized_manifest.job_total_tasks)
+  }
+
+  
+  async launch() {
+    // Only handling launch case where fresh factory is created for each escrow, in future, expand this to mimic python lib
```
uivlis, line 63: Oh, snap! You gotta launch a factory only and only once per hmtoken! And that factory handles all escrows. Please "mimic" this python lib functionality before the merging of this PR.

uivlis (dismissed):
Looks good, try to lint it with a good linter, and most importantly, take care of the most crucial of the issues I've found with your code: do not launch a factory with each job! Also, just curious (probably you mentioned it somewhere, but I'll ask again anyway): have you tested this on the blockchain?

excerebrose:
> Looks good, try to lint it with a good linter, and most importantly, take care of the most crucial of the issues I've found with your code: do not launch a factory with each job! Also, just curious (probably you mentioned it somewhere, but I'll ask again anyway): have you tested this on the blockchain?

Hi @uivlis the issue with factory I'll get fixed and lint it. The reason I made the comment on that was because in the read me it states that:
```
Using only the Manifest and Credentials deploys a new factory to the Ethereum network with the public key of a known Reputation Oracle.
```
And on the first form which was approved we weren't taking in factory address for deployment. Easy fix.

And the reason the hmt token address I used was in your ETHbridge.py, this was the address used
```
HMTOKEN_ADDR = Web3.toChecksumAddress( os.getenv("HMTOKEN_ADDR", "0x4C18A2E51edC5043e9c4B6b0757990A4Ac13797f"))
```

And yes I have, it was failing a couple of times but then I deployed all of the contracts on my own test net, used those addresses and it worked, albeit a bit slower than I'd like.

hmt_escrow_js/ETHInterface.js
excerebrose: From your codebase.  Fixing it
```
HMTOKEN_ADDR = Web3.toChecksumAddress( os.getenv("HMTOKEN_ADDR", "0x4C18A2E51edC5043e9c4B6b0757990A4Ac13797f"))
```

hmt_escrow_js/ETHInterface.js
excerebrose: That's weird, over this the event calls were coming in just fine? Added a note

hmt_escrow_js/ETHInterface.js
excerebrose: Done.. check latest commits

excerebrose:
@uivlis FIxed the changes you mentioned, namely:
1. Move ABIs from hardcoded to generated on the fly using solc-js
2. Fixed new factory address generation only in the case one isnt provided
3. Fixed the form (over in the other repository) to accept factory address.
4. Also noticed some functions werent updated on this repo since most work was done in the other one, so updated & fixed that.

Let me know if you have any questions

gaieges:
Let's merge this and then I can follow up with the npm package before we merge the other repo

gaieges:
@uivlis if you have any other comments in here lets open up another ticket for that

gaieges (approved):

hmt_escrow_js/ETHInterface.js
uivlis: Hmm, I can't see any note, would you point to me to where you added that?

uivlis (commented):
Good job @excerebrose! Don't forget to test this again on the blockchain as a final test. I've caught some ugly variable misspellings that would have certainly caused trouble in a test. Also some error handling suggestions there and we should be good to go.

excerebrose:
> Good job @excerebrose, but where's my linter? I've caught some ugly variable misspellings that would have certainly caused trouble in production. You could have also tested this again on the blockchain as a final test, that would have caught them. Also, some error handling suggestions there and we should be good to go.

Hey @uivlis this is very weird, these weren't there when I tested this directly through the form. Must've been some mix up when moving between repos and wasnt tested independently. Let me do your changes and test manually and not through the form.

PR #231, opened 2020-10-08
Deploy npm alpha package via travis

gaieges:
Stubs for ci / tests


uivlis (approved):

excerebrose:
@gaieges @uivlis I am putting the fixes on this branch, it's the latest

uivlis:
Hey @excerebrose can you fix this CI build? Thanks in advance!

excerebrose:
Think @gaieges will be able to better deploy it.

.travis.yml
```diff
@@ -1,25 +1,46 @@
...
+        - npm run copy-contracts
+        - ./bin/ci
+      deploy:
+        - provider: npm
+          edge: true # opt in to dpl v2
+          access: public
+          tag: alpha
+          on:
+            branch: npm-publish-js
+
```
uivlis, line 41:
```suggestion
          branch: master
          on:
            tags: true
```
@gaieges @excerebrose We should publish only on tag release from master, like here: https://github.com/hCaptcha/hmt-escrow/blob/3fdadc802487e1c1a56eb011fa37866f5546d0ff/.travis.yml#L20-L30

Otherwise, it's gonna publish our package every time we commit to this branch, and we have to change the version every other commit (that's why the latest build fails).

uivlis (changes requested):

.travis.yml
gaieges: I'll flip it back to master but not tying it to tags yet, as the python package is already versioned and they wont move together for now

gaieges:
.5 released yall. I'm updating the docs to describe how to deploy (will be manual for now).  Lets merge this once we're happy with it

PR #234, opened 2020-10-09
Linter added to JS Package

excerebrose:
1. Linter added
2. Fixed vars

Fixes changes mentioned in https://github.com/hCaptcha/hmt-escrow/pull/230

related to issue https://github.com/hCaptcha/hmt-escrow-form/issues/2

### DIFF
--- a/hmt_escrow_js/Job.js
+++ b/hmt_escrow_js/Job.js
@@ -14,7 +14,7 @@ class Job {
     this._manifest_url = manifest_url
     this._factory_addr = factory_addr
 
-    this.serialized_manifest = this._download_manifest()
+    this._serialized_manifest = this._download_manifest()
     this.amount = null
 
     // Validate manifest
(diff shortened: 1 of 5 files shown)


excerebrose:
@uivlis

hmt_escrow_js/Job.js
```diff
@@ -14,7 +14,7 @@ class Job {
     this._manifest_url = manifest_url
     this._factory_addr = factory_addr
 
-    this.serialized_manifest = this._download_manifest()
+    this._serialized_manifest = this._download_manifest()
     this.amount = null
```
uivlis, line 18: Interesting, why does `amount` not have a `_` in front of, just like the others?

excerebrose: stupidity on my part

uivlis: :) Then can I ask, why do all of them have a `_` in front of them? Are they internal in any way?

excerebrose: habit for internals when designing classes, here the usecase is slightly different but I stuck with it

PR #242, opened 2020-10-28
Add WSS support to get_w3() 

### LINKED ISSUE #236: Parametrize the `get_w3` function to choose between `https` or `wss` connection to provider
Recently, I stumbled on a situation where I could only use the `wws` connection to Infura, because events-related RPC methods were only available via Websockets, so I had to build my own web3 object and could not rely on the generic `get_w3` function (which can only use `https`).

This issue requires to parametrize the `get_w3` function (possibly like this: `get_w3("wss") | get_w3("https")`) so that we could have both the `https` and `wss` provider out of the box.

sahzhang:
**Changes**
- edited `get_w3()` to support WSS endpoints set in HMT_ETH_SERVER
     - Technically, this will now support any endpoints that are supported by `web3`'s `AutoProvider`, more below

**Reasoning**
- Fixes #236 

**Motivation**
- Instead of using a parameter, we can utilize a helper function available in `web3.providers.auto` called `load_provider_from_uri` (https://github.com/ethereum/web3.py/blob/master/web3/providers/auto.py#L46-L62) which will return the proper `Provider` to be inserted into `Web3(provider)`. This also adds support for IPC if ever needed

### DIFF
--- a/hmt_escrow/eth_bridge.py
+++ b/hmt_escrow/eth_bridge.py
@@ -5,7 +5,7 @@ import unittest
 
 from solcx import compile_files
 from web3 import Web3
-from web3.providers import HTTPProvider
+from web3.providers.auto import load_provider_from_uri
 from web3.providers.eth_tester import EthereumTesterProvider
 from web3.types import TxReceipt
 from eth_typing import Address, ChecksumAddress, HexAddress, HexStr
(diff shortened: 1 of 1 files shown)


uivlis (approved):

uivlis:
Thanks @sahzhang this is looking good! @gaieges wondering how can we check against the CI tests.

gaieges:
CI issue fixed, looks like this needs to have lint run against it to get it to pass

sahzhang:
Fixed the CI issues! Let me know if there's anything else

gaieges:
Looks great, thanks!

PR #260, opened 2021-04-21
Add honeycomb metrics to blockchain requests

uivlis:
Whenever we submit a tx to the bc, honeycomb will now monitor it, if in the said env there is a `HONEYCOMB_WRITEKEY` and a `HONEYCOMB_SERVICE` sys vars.


gaieges:
We don't want this in escrow, just internally @uivlis

PR #275, opened 2021-10-03
#274 Adds retry logic to setup method in Job class

### LINKED ISSUE #274: High failure rate on setup
Seeing this error

```
INFO::Loading labeling jobs via blockchain Contract.
INFO:hmt_escrow.job:Job's escrow contract deployed to:0x****01
INFO::Contract address: 0x****01
INFO:hmt_escrow.job:Setup failed with main credentials: 0x****aa due to {'code': -32000, 'message': 'replacement transaction underpriced'}.
INFO:hmt_escrow.job:Setup failed with 0x****bb due to {'code': -32000, 'message': 'replacement transaction underpriced'}.
INFO:hmt_escrow.job:Setup failed with 0x****cc and due to {'code': -32000, 'message': 'replacement transaction underpriced'}.
ERROR:hmt_escrow.job:Setup failed with all credentials.
```

rodriguez-facundo:
closes #274

Adds options to retry Job.setup with exponential backoff

### DIFF
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -23,6 +23,7 @@ from hmt_escrow.eth_bridge import (
     get_w3,
     handle_transaction,
 )
+from hmt_escrow import utils
 from hmt_escrow.storage import download, upload
 from basemodels import Manifest
 
@@ -351,7 +352,14 @@ class Job:
         self.manifest_hash = hash_
         return self.status() == Status.Launched and self.balance() == 0
 
-    def setup(self, gas: int = GAS_LIMIT) -> bool:
+    def setup(
+        self,
+        gas: int = GAS_LIMIT,
+        blocking: bool = False,
+        retries: int = 3,
+        delay: int = 5,
+        backoff: int = 2,
+    ) -> bool:
         """Sets the escrow contract to be ready to receive answers from the Recording Oracle.
         The contract needs to be deployed and funded first.
 
@@ -441,6 +449,11 @@ class Job:
             self.manifest_url,
             self.manifest_hash,
         ]
+
+        if not hmt_transferred and blocking:
+            fn = partial(handle_transaction, *func_args, **txn_info)
+            fn.__name__ = hmt_transferred.__name__  # type:ignore
+            hmt_transferred = utils.with_retry(fn, retries, delay, backoff)
 
         try:
             handle_transaction(txn_func, *func_args, **txn_info)
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -452,7 +452,7 @@ class Job:
 
         if not hmt_transferred and blocking:
             fn = partial(handle_transaction, *func_args, **txn_info)
-            fn.__name__ = hmt_transferred.__name__  # type:ignore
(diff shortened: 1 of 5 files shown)


hmt_escrow/job.py
```diff
@@ -442,6 +450,11 @@ def setup(self, gas: int = GAS_LIMIT) -> bool:
             self.manifest_hash,
         ]
 
+        if not hmt_transferred and blocking:
```
vkomodey: this `if` condition is not going to be executed, because in [this line](https://github.com/humanprotocol/hmt-escrow/pull/275/files#diff-41bbe49a252d9374baea7976d28fdf0cbdf6037ec58e38dafe71e78036c8ee74R436) it is already checked(`if not hmt_transferred`)

vkomodey:
So that means that besides `wait_for_transaction_receipt` returns a positive result, the block containing this transaction sometimes is not confirmed by some nodes, and we get the wrong nonce, right?

hmt_escrow/utils.py
```diff
@@ -0,0 +1,47 @@
...
+import logging
+
+
+logger = logging.getLogger("hmt_escrow.job")
+
+
+def with_retry(fn, retries=3, delay=5, backoff=2):
+    """ Retry a function
+
+    Mainly for BC transactions. Uses expnential backoff.
```
vkomodey: what is BC transaction here?

hmt_escrow/job.py
vkomodey: I think, here you meant `if hmt_transferred and blocking`

vkomodey (changes requested):

rodriguez-facundo:
> So that means that besides `wait_for_transaction_receipt` returns a positive result, the block containing this transaction sometimes is not confirmed by some nodes, and we get the wrong nonce, right?


I would say so. the method get nonce I think returns the nonce with confirmed block. Because we have a list of gas payers and we fail reliably with all of them. this will give the ability to wait until the block is confirmed which will give us the next nonce.

Will fix test...

hmt_escrow/job.py
rodriguez-facundo: if hmt_transfered, then I don't want to execute the code. Yeah, I agree, negations are always troublesome. But I don't want to introduce too many changes.

hmt_escrow/utils.py
rodriguez-facundo: blockchain. will work on a better description for this method.

hmt_escrow/job.py
vkomodey: here I see the next code:

```
# 1
if not hmt_transferred:
      return False
...
# 2 
if not hmt_transferred and blocking:
     # .. your code
```

and the `# 2` will not be executed because of the return statement in `# 1`

maybe I don't see something in here:)

rodriguez-facundo: yes, that means that the contract was never created, so it is safe to skip the setup step

rodriguez-facundo: bah, you are right. the flag is contract_is_setup not hmt_transferred.

rodriguez-facundo: Updating...

vkomodey (approved):

PR #279, opened 2021-10-27
remove Trinity dependency/move crypto code to hmt-escrow

### LINKED ISSUE #272: Remove dependency on the `trinity` package
We have the package only for its dependencies, not for the package itself. What we actually use is `web3==5.12.0` and `py-evm==0.3.0a18`.
Since the `trinity` package is not maintained and its repository archived, there are no benefits of keep using it and it can be replaced with the real dependencies.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200560623420491/1201175417135913) by [Unito](https://www.unito.io)

adam-mrozik:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200560623420491/1201278021276808) by [Unito](https://www.unito.io)
Closes #272


Pipfile
```diff
@@ -6,14 +6,18 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-trinity = {git = "https://github.com/ethereum/trinity.git", ref = "master"}
-py-solc-x = {git = "https://github.com/iamdefinitelyahuman/py-solc-x.git", ref = "master"}
 boto3 = "*"
 hmt-basemodels = "*"
-Sphinx = {git = "https://github.com/sphinx-doc/sphinx.git", ref = "master"}
+web3 = "==5.12.0"
```
e271828-: Is there a reason to pin this to a point release?

hmt_escrow/crypto.py
```diff
@@ -0,0 +1,168 @@
+import os
```
e271828-, line 24: Where did this file come from? Please document source + license at the top. I'd guess it's MIT from a quick search on the strings, so looks fine, but needs to be recorded.

adam-mrozik: Added license + source

Pipfile
adam-mrozik: Hmm, it was suggested in #272 , but tests work anyway, so I will leave it unpinned

Pipfile
```diff
@@ -6,14 +6,15 @@ verify_ssl = true
...
+py-solc-x = "*"
+sphinx = "*"
+web3 = "*"
 
 [requires]
 python_version = "3.8"
 
 [pipenv]
-allow_prereleases = true
+allow_prereleases = true
```
spirit-q2:
```suggestion
allow_prereleases = true

```

spirit-q2 (commented):
a small comment :)

adam-mrozik:
@e271828- 
Please advise me if this is correct and if can be approved :)

e271828- (approved):

PR #281, opened 2021-10-30
Fix version of `web3` to resolve BC breaks

spirit-q2:
we have to lock `web3` at `v5.20.0` since after that version they switched to `websockets` of `v9.1` which brings BC break (comparing to `v8.1`) to our code by deprecating some classes that we use. I.e. `websockets.protocol` became `websockets.legacy.protocol`.

Related to #272 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200560623420491/1201298108242295) by [Unito](https://www.unito.io)

### DIFF
--- a/hmt_escrow/storage.py
+++ b/hmt_escrow/storage.py
@@ -172,7 +172,12 @@ def _decrypt(private_key: bytes, msg: bytes) -> str:
 
     """
     priv_key = keys.PrivateKey(codecs.decode(private_key, "hex"))
-    e = crypto.decrypt(msg, priv_key, shared_mac_data=SHARED_MAC_DATA)
+    shared_mac_data = (
+        SHARED_MAC_DATA
+        if isinstance(SHARED_MAC_DATA, bytes)
+        else SHARED_MAC_DATA.encode("ascii")
+    )
+    e = crypto.decrypt(msg, priv_key, shared_mac_data=shared_mac_data)
     return e.decode("utf-8")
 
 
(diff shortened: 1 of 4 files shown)


hmt_escrow/storage.py
```diff
@@ -196,7 +201,12 @@ def _encrypt(public_key: bytes, msg: str) -> bytes:
     """
     pub_key = keys.PublicKey(codecs.decode(public_key, "hex"))
     msg_bytes = msg.encode("utf-8")
-    return crypto.encrypt(msg_bytes, pub_key, shared_mac_data=SHARED_MAC_DATA)
+    shared_mac_data = (
+        SHARED_MAC_DATA
+        if isinstance(SHARED_MAC_DATA, bytes)
+        else SHARED_MAC_DATA.encode("ascii")
+    )
```
rustyro: @spirit-q2 can this be done at the top of the file, so it can be reused here and in the reference above w/o the repetition?

e271828-:
Why don't we simply do a compatible import, i.e.

```
try:
    from x.websockets import protocol as websocketsprotocol
except:
    from x.legacy.websockets import protocol as websocketsprotocol
```

?

spirit-q2:
@e271828-, there might be other incompatibilities as well. Although, there is a list of BC breaks: https://websockets.readthedocs.io/en/latest/project/changelog.html#id11 :thinking: 
I can do some investigation on the usage of code that was moved/deprecated if you think it's worth spending time on it.

hmt_escrow/storage.py
spirit-q2: found nice solution :slightly_smiling_face:

spirit-q2: seems to be working :tada:

e271828-:
> @e271828-, there might be other incompatibilities as well. Although, there is a list of BC breaks: https://websockets.readthedocs.io/en/latest/project/changelog.html#id11 🤔 I can do some investigation on the usage of code that was moved/deprecated if you think it's worth spending time on it.

Seems like we should at minimum switch to use the new refs so we can upgrade. Pinning to an arbitrary stale version with no ability to move forward short of code changes will make any security-related upgrades challenging.

spirit-q2:
Makes sense 👍 will do. I like your approach for upgrading the versions :)

spirit-q2:
@e271828- we are done here, since the `hmt-escrow` don't use the `websockets` package directly. BC breaks should be handled on our side. Please review/publish if all fine in the PR

e271828- (approved):

PR #302, opened 2022-03-29
use secondary gas payers if main one is low on hmt

### LINKED ISSUE #301: Validate enough HMT in gas payer before running job.setup
If a gas payer does not have enough HMT to fund an escrow, we don't want to run job.setup() since on that gas payer since we know the transaction will be reverted.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200560623420491/1202045579698114) by [Unito](https://www.unito.io)

rodriguez-facundo:
closes #301 

When the main gas payer is low in HMT, the transaction is accepted and then reverted, causing jobs not to be deployed.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200560623420491/1202045638243182) by [Unito](https://www.unito.io)

### DIFF
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -1845,6 +1845,7 @@ class JobTestCase(unittest.TestCase):
             1000000,
         )
 
+
 if __name__ == "__main__":
     from test_manifest import manifest
 
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -25,6 +25,7 @@ from hmt_escrow.eth_bridge import (
     get_w3,
     handle_transaction_with_retry,
     Retry,
+    HMTOKEN_ADDR
 )
 from hmt_escrow import utils
 from hmt_escrow.storage import download, upload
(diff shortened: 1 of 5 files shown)


hmt_escrow/job.py
```diff
@@ -434,13 +435,18 @@ def setup(self, gas: int = GAS_LIMIT, sender: str = None) -> bool:
...
-        try:
-            handle_transaction_with_retry(txn_func, self.retry, *func_args, **txn_info)
-            hmt_transferred = True
-        except Exception as e:
-            LOG.debug(
-                f"{txn_event} failed with main credentials: {self.gas_payer}, {self.gas_payer_priv} due to {e}. Using secondary ones..."
-            )
+        balance = utils.get_hmt_balance(self.gas_payer, HMTOKEN_ADDR, get_w3())
+
+        if not balance or balance > hmt_amount:
```
posix4e: Maybe a comment explaining what’s up

hmt_escrow/utils.py
```diff
@@ -46,3 +46,30 @@ def with_retry(fn, retries=3, delay=5, backoff=2):
...
+                "inputs": [{"name": "_owner", "type": "address"}],
+                "name": "balanceOf",
+                "outputs": [{"name": "balance", "type": "uint256"}],
+                "type": "function",
+            }
+        ]
+        contract = w3.eth.contract(abi=abi, address=token_addr)
+        return contract.functions.balanceOf(wallet_addr).call()
+    except Exception as e:
+        logger.warning(f"Can't get HMT Balance: {e}")
```
posix4e: Stupid question? Why not let it fail?

rodriguez-facundo: got used to adding a try/exception to everything that runs in the path of the request. Not needed here I think. removing it.

vkomodey (approved):

PR #315, opened 2022-04-27
Implement blockchain multi-network

alidzm:


posix4e (approved):
Seems
Fine but how to test?

rodriguez-facundo (approved):
lgtm

alidzm:
> Seems Fine but how to test?

I think copy library directly into repo and test on yellow account.
@rodriguez-facundo may be do you have some ideas how to test it?

posix4e (approved):

vkomodey:
Hey, @alidzm ! Could you please add some description here which describes the motivation of the change?
I mean, previously we configured hmt-escrow through the environment variables. Right now we are able to do it with function arguments as well. So if we want to use a different blockchain for some functions, we need to be sure that all other environment variables or data derived from these variables is in sync for a new blockchain

For example, when I configure library for the Ethereum network - I setting up the HMT address, kv store, appropriate gas variables, gas payers and etc. But when I want to switch to the Polygon - HMT, KVStore, gas values are different(gas payer can be the same in theory). So I can't simply change the HMT_ETH_SERVER, I need to change different values as well

I like the approach with function variables in general - it gives us more flexibility. But in conjunction with env variables we will have 2 sources of truth, and we need constantly check, whether they are in sync. I'm thinking about approach, where you can initiate several instances for different networks and give all required parameters at once to some constructor or whatever form


```python
polygon_hmt_escrow = hmt_escrow(rpc_api="https://polygon-rpc", hmt_addr="0x0", kv_addr="0x0", gas_limit=1)
ethereum_hmt_escrow = hmt_escrow(rpc_api="https://polygon-rpc", hmt_addr="0x0", kv_addr="0x0", gas_limit=1)
```

what do you think?

BTW: you need to sign the CLA(check the first comment in the PR)

e271828-:
@alidzm if you find yourself adding the same func arg to 20 funcs in the same file, it should likely be a class instantiation arg that's passed/set only once.

hmt_escrow/eth_bridge.py
```diff
@@ -136,7 +147,7 @@ def handle_transaction(txn_func, *args, **kwargs) -> TxReceipt:
 
 
 def handle_transaction_with_retry(
-    txn_func, retry=Retry(), *args, **kwargs
+    txn_func, retry=Retry(), *args, hmt_server_addr: str = None, **kwargs
```
igervasiychuk: I'm not sure we need hmt_server_addr here, as we already have **kwargs, and pass them on later on.

hmt_escrow/eth_bridge.py
```diff
@@ -94,7 +101,9 @@ def get_w3() -> Web3:
     return w3
 
 
-def handle_transaction(txn_func, *args, **kwargs) -> TxReceipt:
+def handle_transaction(
```
igervasiychuk: Can we use kwargs to pass the hmt_server_addr?

hmt_escrow/job.py
```diff
@@ -417,7 +421,10 @@ def setup(self, gas: int = GAS_LIMIT, sender: str = None) -> bool:
         reputation_oracle = str(self.serialized_manifest["reputation_oracle_addr"])
         recording_oracle = str(self.serialized_manifest["recording_oracle_addr"])
         hmt_amount = int(self.amount * 10 ** 18)
-        hmtoken_contract = get_hmtoken()
+        hmtoken_address = (
```
igervasiychuk: It might be better to init self.hmtoken_addr in __init__, and just use self.hmtoken_addr in get_hmtoken in the line below.

hmt_escrow/eth_bridge.py
```diff
@@ -77,13 +77,20 @@ def get_w3() -> Web3:
...
 
     """
-    endpoint = os.getenv("HMT_ETH_SERVER", "http://localhost:8545")
+    endpoint = None
+
+    if hmt_server_addr:
+        endpoint = hmt_server_addr
+
     if not endpoint:
-        LOG.error("Using EthereumTesterProvider as we have no HMT_ETH_SERVER")
```
igervasiychuk, line 85: This removes error logging, in case where the HMT_ETH_SERVER env variable is defined and is "" (empty string). If it's accidental, might be better to return it back.

hmt_escrow/job.py
```diff
@@ -1830,7 +1905,15 @@ def test__raffle_txn_retry(self):
...
-                [call(txn_mock, gas_payer="1", gas_payer_priv="11", gas=6700000)],
+                [
+                    call(
+                        txn_mock,
+                        gas_payer="1",
+                        gas_payer_priv="11",
+                        gas=6700000,
+                        hmt_server_addr=None,
+                    )
+                ],
```
igervasiychuk: New functionality has been added: e.g. starting new Jobs on two different networks. It might make sense to add dedicated tests for this. It will likely require changing get_w3 function in eth_bridge.py .

igervasiychuk (approved):

hmt_escrow/job.py
```diff
@@ -1804,7 +1823,13 @@ def test__raffle_txn_retry(self):
             self.assertEqual(
                 handler_mock.call_args_list,
                 [
-                    call(txn_mock, gas_payer="1", gas_payer_priv="11", gas=6700000)
+                    call(
+                        txn_mock,
+                        gas_payer="1",
+                        gas_payer_priv="11",
+                        gas=6700000,
```
posix4e: Why 67M

hmt_escrow/job.py
alidzm: done

hmt_escrow/eth_bridge.py
alidzm: done

hmt_escrow/job.py
alidzm: @posix4e this is just test-case which checks that handler is called with right parameters. I didn't changed values here, just added one `hmt_server_addr`

hmt_escrow/eth_bridge.py
alidzm: done

hmt_escrow/eth_bridge.py
alidzm: done

vkomodey:
@alidzm
hey! any thoughts about [my previous comment](https://github.com/humanprotocol/hmt-escrow/pull/315#issuecomment-1114754185)?

alidzm:
> @alidzm hey! any thoughts about [my previous comment](https://github.com/humanprotocol/hmt-escrow/pull/315#issuecomment-1114754185)?

> 

I think that current implementation is backward compatible and we can manage connections using env variable that provides outside from any application. Only one question is in `GAS_LIMIT`. Will check how to implement it with backward compatibility.

hmt_escrow/job.py
```diff
@@ -315,14 +321,14 @@ def launch(self, pub_key: bytes) -> bool:
...
         LOG.info("Job's escrow contract deployed to:{}".format(job_addr))
-        self.job_contract = get_escrow(job_addr)
+        self.job_contract = get_escrow(job_addr, self.hmt_server_addr)
 
         (hash_, manifest_url) = upload(self.serialized_manifest, pub_key)
         self.manifest_url = manifest_url
         self.manifest_hash = hash_
         return self.status() == Status.Launched and self.balance() == 0
 
-    def setup(self, gas: int = GAS_LIMIT, sender: str = None) -> bool:
```
vkomodey, line 325: This is not a backward-compatible API change. Due to this, a major part of the package version should be changed

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hmt-escrow",
-  "version": "0.11.0",
+  "version": "0.12.0",
```
vkomodey: You've changed a patch version, while we have a breaking change here. We need to change a major version. https://semver.org/

alidzm: Do you mean `1.0.0` version?

vkomodey: yep

PR #327, opened 2022-06-13
Adding support to decide whether final results must be encrypted

hugoseabra-imachines:
- Added new crypto module with better encrypting definitions, façade and tests dedicated to it;
- Added dedicated folder to run tests;
- Added option to support decision whether data being uploaded to storage must be encrypted or plain;
- Added flag to decide whether final results in Job Bulk payouts must be encrypted or plain;


test/Escrow.test.js
```diff
@@ -80,7 +80,7 @@ contract('Escrow', (accounts) => {
 
     it('fails when aborting with different address', async () => {
       try {
-        tx = await Escrow.abort({ from: externalAddress });
```
igervasiychuk, line 83: this doesn't look correct, tx is used on the next line. 
and I think this test won't break if someone were to remove the "trusted" modifier on the "abort" method in the escrow contract

bin/wait_then_run
```diff
@@ -30,7 +30,5 @@ export PYTHONPATH=$(pwd)
 if [ ! -z $* ]; then
   $*
 else
-  python3 hmt_escrow/job.py
-  python3 hmt_escrow/storage.py
-  python3 hmt_escrow/eth_bridge.py
+  python3 -m unittest discover -s test/hmt_escrow
```
igervasiychuk, line 31: I would've created two PRs, one with the refactoring, and one with the two-liner for the "Adding support to decide whether final results must be encrypted".

igervasiychuk (changes requested):

bin/wait_then_run
hugoseabra-imachines: Great point. I should've done that, indeed.

Would you like me to separate the both implementations in different PR?

hugoseabra-imachines: @igervasiychuk , I will close this PR in favor of PR split to other 3:
- https://github.com/humanprotocol/hmt-escrow/pull/328
- https://github.com/humanprotocol/hmt-escrow/pull/329

PR #329, opened 2022-06-13
Crypto module refactored with tests

hugoseabra-imachines:
Reorganized tests to a dedicated repository, as it was causing cyclic dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "hmt-escrow",
-  "version": "0.10.0",
+  "version": "0.10.2",
   "lockfileVersion": 1,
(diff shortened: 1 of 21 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hmt-escrow",
-  "version": "0.9.10",
+  "version": "0.10.0",
```
vkomodey: should be 0.10.2 as well

vkomodey (changes requested):
All is good. Just change the version in the package-lock.json and we are good to go

vkomodey (approved):

PR #330, opened 2022-06-13
Added final results encryption option

hugoseabra-imachines:
Added flag to decide whether final results in Job Bulk payouts must be encrypted or plain;


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hmt-escrow",
-  "version": "0.10.2",
+  "version": "0.10.3",
```
vkomodey: This is a compatible API change. The version should be 0.11.0

vkomodey (changes requested):
lgtm

vkomodey:
And please, squash all your commits into 2 or 3 or 4 commits. Thanks

package.json
hugoseabra-imachines: @vkomodey , I kept usage compatibility with upload by adding optional flag parameter. I really believe it is a minor version as it does not break other consumers. Don't you think we should keep minor change in this way?

vkomodey: According to [semver](https://semver.org/) rules, we should change the PATCH number:

> * MAJOR version when you make incompatible API changes,
> * MINOR version when you add functionality in a backwards compatible manner, and
> * PATCH version when you make backwards compatible bug fixes.

hugoseabra-imachines: Oh sure! Doing it!

hugoseabra-imachines:
> And please, squash all your commits into 2 or 3 or 4 commits. Thanks

Due to the sequence of commits where there are commits that are before the merge with `master`, when squashing the commits, the changes before `master` is lost.

Is it possible to merge this without squashing?

The sequence of commits. Just note that there commits of this branch before `master` merge:
image

vkomodey:
@hugoseabra-imachines 
Please, never use `merge ${stable branch name, i.e. master}` to your dev branches. To keep your commit history clean - please, always use the `rebase` command. 

`git pull origin master --rebase`

It will allow you to keep master commits behind new commits from your branch.

As for this branch, you can check out a new branch(from master) and do a cherry-pick of all required commits into your history. Or simple copy&paste content from this PR into the new branch will do the work. I would say it is the easiest and quickest option here

igervasiychuk:
will this do? @vkomodey  https://github.com/humanprotocol/hmt-escrow/pull/336

vkomodey:
Duplicated and merged in #336

PR #343, opened 2022-07-07
S3 full qualified URL for public final results

hugoseabra-imachines:
When encyrption is off for final results, the URL sent in transaction must be persisted as full qualified URL and final results must be stored on bucket which allows public acess to its files.

- [x] Compatibility with `0.12.0` maintained;
- [x] Tests added to final results methods;
- [x] Tests added to handle bulk payout with encryption on/off;
- [x] Enhanced/Decoupled storage module function to better handle bucket decisions and retrieving key and url;

@alidzm asked to cherry-pick a small change from him to the PR:
- [x] Added gas limit parameters, otherwise use GAS_LIMIT env var;


vkomodey:
Hey guys. I don't understand the idea of:
> When encyrption is off for final results, the URL sent in transaction must be persisted as full qualified URL and final results must be stored on bucket which allows public acess to its files.

This is something not obvious from the library client's perspective. We might need to store only S3 identifier **only** if the client decides so. But I can see cases when we might need a full URL for encrypted results: for example, if we want to give access to final results to everyone who has the appropriate private key, but not all parties have a secret key from the S3(or any other storage).

Instead of the code like:

```
# Plain data will be publicly accessible
is_public_url = encrypt_final_results is False
url = get_public_bucket_url(url) if is_public_url else URL
```

let's give a user ability to choose, whether he wants to store public accessible URL by himself. For example, by adding an additional environment variable to the list: `IS_ESCROW_MANIFEST_PUBLIC_URL` which equal false by default(to add no breaking changes to the lib)

bin/wait_then_run
```diff
@@ -24,6 +24,7 @@ until curl --silent ipfs:5001; do
 done
 
 mkdir -p minio/escrow-results
+mkdir -p minio/escrow-public-results
```
vkomodey, line 27: any reason to have an additional bucket here? I think, instead of adding additional logic and `if's` it is better to have 1 bucket and a bunch of configuration that controls the behavior: like "enable encryption", "show the full URL" and etc

the code will be much more simpler and straightforward, what do you think?

hugoseabra-imachines: Hi @vkomodey , 

the request was to support publicly exposed data, when encryption is off, we'll show the full qualified URL - and final results to be persisted in blockchain plainly. Also, the full URL will be publicly exposed to user to download the files whenever he wants.

However, to support public files in bucket in S3, according to [this article](https://aws.amazon.com/premiumsupport/knowledge-center/read-access-objects-s3-bucket/), we decrease bucket security by deciding that some files are public and others might not be. To do that we'll have to configure the bucket to support public files and ACL will determine that, which makes the bucket vulnerable with possibility to expose non-public chunk data. So, by determining that the information is public and chunk file is downloadable, it is reasonable (and secure) to send data to bucket that supports public exposure of its files stored.

So, I believe we should move the discussion to decide the best approach. And, I completely agree, using one single bucket would make everything simpler and code less complex.

@igervasiychuk , can you please lead the business strategy to follow up?

igervasiychuk: @vkomodey from my perspective, there's this little extra complexity in configuration. But it's essentially solved by copy-pasting the s3 creds if you'd like the same bucket. 

On the other hand, if you would like a better security at some point, by having another bucket, you'd be out of luck, and need dev time. 

But the decision, I think, can be yours here.

igervasiychuk:
> let's give a user ability to choose, whether he wants to store public accessible URL by himself. For example, by adding an additional environment variable to the list: `IS_ESCROW_MANIFEST_PUBLIC_URL` which equal false by default(to add no breaking changes to the lib)

I agree, this is a nice, more granular level of control. But do you think it also should be included in the manifest, and not as an env var?  

Because if it's an env var, you'd need a separate installation every time you need to use this feature?

vkomodey:
>I agree, this is a nice, more granular level of control. But do you think it also should be included in the manifest, and not as an env var?

Yes. Agree. Env variable is not a good option here. The crucial moment is that some parties(like recording or reputation oracle) can get the data from a public place(S3 or IPFS or any other storage) and decrypt it using their private keys. But we have a strong policy "access is public => data is not encrypted", "access is private => data is encrypted". So this might be your case guys, but we should allow users to choose how to store the data

bin/wait_then_run
vkomodey: Agree. Let's have several buckets

e271828-:
@vkomodey as pointed out above, this is definitely not appropriate for an env var. Also, there may be some confusion here about which URL is public.

In this use case, the **results** URL is the only one that must be public.

The **manifest** URL can be public or not, but there is a good argument for not making it public at launch, i.e. leaving it encrypted: this prevents some types of front-running attacks.

vkomodey (approved):

vkomodey:
@hugoseabra-imachines please, update the version and we are gtg

PR #348, opened 2022-08-12
Read Transfer event for transfer transactions

alidzm:
The change fixes issue related to transfer tokens transaction.
Sometimes there is a valid transaction without event. In this case flag `hmt_transferred` will be `True`, but actually this is wrong `hmt_transferred` should be `False`. Changed code checks Transfer event of transaction and if it exists switches `hmt_transferred` to `True`.
Moreover `self.balance()` may also return wrong value if one of blockchain node still doesn't have information about transaction, code takes balance from `tx_receipt`.

### DIFF
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -3,14 +3,14 @@ import logging
 import os
 from decimal import Decimal
 from enum import Enum
-from typing import Dict, List, Tuple, Optional, Any
+from typing import Dict, List, Tuple, Optional, Any, TypedDict
 
 from basemodels import Manifest
 from eth_keys import keys
 from eth_utils import decode_hex
 from web3 import Web3
 from web3.contract import Contract
-from web3.types import Wei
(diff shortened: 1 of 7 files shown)


Dzeranov:
@alidzm please, provide a description to this PR.

igervasiychuk (approved):
I wish we didn't pin the dependencies

hugoseabra-imachines:
> I wish we didn't pin the dependencies

hugoseabra-imachines:
> I wish we didn't pin the dependencies

@alidzm , we you worked around `cryptography` package error about the SO by adding a different ubuntu image in docker, we may keep version as `*`, as version `cryptography` v37 is announced to be long-term.

alidzm:
> @alidzm please, provide a description to this PR.

@Dzeranov Done

hugoseabra-imachines (approved):

Dzeranov (approved):

PR #350, opened 2022-08-15
Environment fix

hugoseabra-imachines:
When building local image, many broken npm and python packages due to `ubuntu:focal` packages.

When updating to `ubuntu:jammy`, the following changes were needed:

[x] `python` version to `3.10`;
[x] downgrade `npm` to version `6` to run `gyp` correctly;
[x] update `package.json` for vulnerability fixes (good to have);
[x] added `python-is-python3` package as `python` command is not default in new Ubuntu/Debian versions;

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,10 +6,10 @@ ENV PYTHONUNBUFFERED True
 WORKDIR /work
 
 RUN apt-get update -y && \
-    apt-get install -y automake bash black build-essential curl git jq libffi-dev libgmp-dev libtool mypy nodejs npm \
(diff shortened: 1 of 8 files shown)


Dzeranov (approved):

igervasiychuk (approved):
lgtm, seemingly random applying of formatting seems strange

hugoseabra-imachines:
> lgtm, seemingly random applying of formatting seems strange

I had to remove `mypy` from apt-get and added it with pip. It's version updated with a completely different behavior at linting.

PR #353, opened 2022-08-23
Skale implementation

AzizBardi:
Skale implementation


foufrix:
Hi @AzizBardi 
This commit seems to have way too much change, only truffle.js and subgraph should have changed, like this PR :https://github.com/humanprotocol/hmt-escrow/pull/354/

AzizBardi:
Hello@foufrix 

Truffle.js package.js and docker files were changed to make the project work.
These changes were necessary to make the project work.

foufrix:
You've pushed a change with 104 files changes, where there is only 2 needed, you need to clean this PR or it will be rejected.

foufrix (changes requested):
Too much files changed, 
The commit need to be cleaned

PR #355, opened 2022-08-26
Add staking/slashing features

ghost:
- Escrow creation process support staking. So only parties that are in the staking network are allowed to launch escrows from the given factory. 
- Contracts support slashing. Users can be a member of "validators" committee which can vote against bad actors. User can become a validator by staking token as much as validator's limit.
- There is a pool of tokens available for rewards. All slashed tokens go to this pool. IERC20 interface was used as a staking token interface.


e271828- (commented):
This kind of code needs a lot of comments. For example:

- What does each function do?
- What are the expected constraints on function inputs?

posix4e:
Thanks for the POC, closing now. Happy to chat more online

PR #360, opened 2022-09-07
Added counter to the Launched event

0xVoronov:
Updated **Launched** event in Escrow Factory contract. Counter has been added to the event. The counter will be used for historical data in the dashboard.


m00n620 (approved):
LGTM

PR #368, opened 2022-09-23
Get escrow address from tx_receipt

alidzm:
After creating contract to get `escrow_address` there used `self._last_escrow_addr()` what is wrong, because if someone run creation contracts in parallel `self._last_escrow_addr()` can return the same address for two contracts, that's wrong. The PR contains rework of wrong implementation: `escrow_address` takes from `tx_receipt` of create contract transaction.

### DIFF
--- a/hmt_escrow/job.py
+++ b/hmt_escrow/job.py
@@ -335,12 +335,14 @@ class Job:
         # Use factory to deploy a new escrow contract.
         trusted_handlers = [addr for addr, priv_key in self.multi_credentials]
 
-        txn_success = self._create_escrow(trusted_handlers)
+        txn = self._create_escrow(trusted_handlers)
 
-        if not txn_success:
+        if not txn["txn_succeeded"]:
             raise Exception("Unable to create escrow")
 
-        job_addr = self._last_escrow_addr()
+        tx_receipt = txn["tx_receipt"]
+        events = self.factory_contract.events.Launched().processReceipt(tx_receipt)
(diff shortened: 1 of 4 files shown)


hmt_escrow/job.py
```diff
@@ -335,12 +335,14 @@ def launch(self, pub_key: bytes) -> bool:
...
-        txn_success = self._create_escrow(trusted_handlers)
+        txn = self._create_escrow(trusted_handlers)
 
-        if not txn_success:
+        if not txn["txn_succeeded"]:
             raise Exception("Unable to create escrow")
 
-        job_addr = self._last_escrow_addr()
+        tx_receipt = txn["tx_receipt"]
+        events = self.factory_contract.events.Launched().processReceipt(tx_receipt)
```
igervasiychuk, line 344: I think this will throw an exception because on polygon the logs contain 2 events, and one of them isn't "Launched" event

igervasiychuk (changes requested):

hmt_escrow/job.py
igervasiychuk: Could you possibly check that?

igervasiychuk (approved):

hmt_escrow/job.py
alidzm: Checked, we have one event for currently supported networks: polygon, polygon-mumbai, rinkeby

Dzeranov (approved):

PR #371, opened 2022-09-30
Fix store artifact to public or private bucket

alidzm:
Fix incorrect logic of deploying encrypted and unencrypted data to public/private storage.

### DIFF
--- a/hmt_escrow/storage.py
+++ b/hmt_escrow/storage.py
@@ -29,6 +29,13 @@ ESCROW_PUBLIC_BUCKETNAME = os.getenv(
 ESCROW_AWS_ACCESS_KEY_ID = os.getenv("ESCROW_AWS_ACCESS_KEY_ID", "minio")
 ESCROW_AWS_SECRET_ACCESS_KEY = os.getenv("ESCROW_AWS_SECRET_ACCESS_KEY", "minio123")
 
+ESCROW_RESULTS_AWS_S3_ACCESS_KEY_ID = os.getenv(
+    "ESCROW_RESULTS_AWS_S3_ACCESS_KEY_ID", ""
+)
+ESCROW_RESULTS_AWS_S3_SECRET_ACCESS_KEY = os.getenv(
+    "ESCROW_RESULTS_AWS_S3_SECRET_ACCESS_KEY", ""
+)
+
 ESCROW_AWS_REGION = os.getenv("ESCROW_AWS_REGION", "us-west-2")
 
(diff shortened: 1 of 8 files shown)


hmt_escrow/storage.py
```diff
@@ -203,20 +213,18 @@ def upload(
...
 
-    if encrypt_data is True:
-        # If encryption is enabled, the bucket is private and data is encrypted
-        bucket_kwargs.update({"Body": crypto.encrypt(public_key, artifact)})
-    else:
-        # If encryption is off, file will be publicly readable
-        bucket_kwargs.update({"ACL": "public-read", "Body": content})
-
-    BOTO3_CLIENT = _connect_s3()
+    BOTO3_CLIENT = _connect_s3(use_public_bucket)
```
igervasiychuk: could you make a little change to PEP8's boto3_client (lowercase)?

hmt_escrow/storage.py
```diff
@@ -29,6 +29,9 @@
 ESCROW_AWS_ACCESS_KEY_ID = os.getenv("ESCROW_AWS_ACCESS_KEY_ID", "minio")
 ESCROW_AWS_SECRET_ACCESS_KEY = os.getenv("ESCROW_AWS_SECRET_ACCESS_KEY", "minio123")
 
+ESCROW_AWS_S3_ACCESS_KEY_ID = os.getenv("ESCROW_AWS_S3_ACCESS_KEY_ID", "")
```
igervasiychuk: AWS_S3 is implementation details. can it be named something like "ESCROW_PUBLIC_BUCKET_..."? 

e.g. like it's used in ESCROW_PUBLIC_RESULTS_URL

igervasiychuk (approved):

hmt_escrow/storage.py
alidzm: No, it's a user with programmatic access to S3 with read and write permissions, it will be used for write/read to/from public and private (in future) buckets.

alidzm: May be `ESCROW_RESULTS_AWS_S3_ACCESS_KEY_ID` ?

tikibu: ok

hmt_escrow/storage.py
alidzm: done

hmt_escrow/storage.py
igervasiychuk: you are right. you don't have to change anything here. ty

alidzm: Changed to `ESCROW_RESULTS_AWS_S3_...` for more transparency

Dzeranov (approved):

PR #372, opened 2022-10-03
Add holders indexation

Charlie85270:
- Index all maintain all token holders with transfer Events 
- Get the total count of holders in real time
- Get the list and balance of all holders  in real time


hmt_subgraph/src/mapping/hm-token.ts
```diff
@@ -149,3 +158,42 @@ export function handleBulkApproval(event: BulkApproval): void {
...
+
+  if (
+    (isNew && !holder.balance.isZero()) ||
+    (!isNew && balanceBeforeTransfer.isZero())
+  ) {
+    count = 1;
+  } else if (holder.balance.isZero()) {
+    count = -1;
+    store.remove("Holder", id);
+  }
```
portuu3: shouldn't be possible to have a holder which is not new and his old balance was 0.
I think this would be a simpler solution:
```suggestion
    if (isNew && !holder.balance.isZero()) {
    count = 1;
  } else if (!isNew && holder.balance.isZero()) {
    count = -1;
    store.remove("Holder", id);
  }
```

hmt_subgraph/schema.graphql
```diff
@@ -94,6 +94,13 @@ type HMTokenStatistics @entity {
   totalBulkTransferEventCount: BigInt!
   totalValueTransfered: BigInt!
   token: Bytes! # token address
+  holders: Int!
```
portuu3: It is not highly likely, but at some point a BigInt might be necessary

Charlie85270: Updated

hmt_subgraph/src/mapping/hm-token.ts
Charlie85270: I updated the code and removed `store.remove("Holder", id);` so we can keep the history of hodler even if their balance is 0

portuu3: The token contract allows to transfer amount 0. So in this case if the destination address balance was 0 before the value returned would be -1.

You can verify the error using this test in hmt.test.ts:
```describe("HMToken entity", () => {
  test("Should properly calculate holders in statistics", () => {
    let transfer1 = createTransferEvent(
      "0x0000000000000000000000000000000000000000",
      "0x92a2eEF7Ff696BCef98957a0189872680600a959",
      10,
      BigInt.fromI32(10)
    );

    let transfer2 = createTransferEvent(
      "0x92a2eEF7Ff696BCef98957a0189872680600a959",
      "0xD979105297fB0eee83F7433fC09279cb5B94fFC6",
      0,
      BigInt.fromI32(10)
    );

    handleTransfer(transfer1);
    handleTransfer(transfer2);

    assert.fieldEquals(
      "HMTokenStatistics",
      HMT_STATISTICS_ENTITY_ID,
      "holders",
      "1"
    );

    let transfer3 = createTransferEvent(
      "0x0000000000000000000000000000000000000000",
      "0xD979105297fB0eee83F7433fC09279cb5B94fFC6",
      10,
      BigInt.fromI32(10)
    );
    
    handleTransfer(transfer3);

    assert.fieldEquals(
      "HMTokenStatistics",
      HMT_STATISTICS_ENTITY_ID,
      "holders",
      "2"
    );

    let transfer4 = createTransferEvent(
      "0x92a2eEF7Ff696BCef98957a0189872680600a959",
      "0xD979105297fB0eee83F7433fC09279cb5B94fFC6",
      10,
      BigInt.fromI32(10)
    );

    handleTransfer(transfer4);
   
    assert.fieldEquals(
      "HMTokenStatistics",
      HMT_STATISTICS_ENTITY_ID,
      "holders",
      "1"
    );

    clearStore();
  });
});
```

_*This test should be added when fixed._

Solution:
```
if (balanceBeforeTransfer.isZero() && !holder.balance.isZero()) {
    count = 1;
  } else if (!balanceBeforeTransfer.isZero() && holder.balance.isZero()) {
    count = -1;
  }
```
And remove isNew variable (unused)

portuu3 (changes requested):
Fix updateHolders function

hmt_subgraph/src/mapping/hm-token.ts
Charlie85270: Indeed I had not taken into account that the contract allows to transfer amount 0, I updated the the mapper and added the test. Thanks you !

m00n620 (approved):

portuu3 (approved):

Charlie85270:
Screenshot for the example how it's display on the dashboard master branch :
image

Query on the subgraph : 
image

PR #397, opened 2022-10-21
Fortune merge

### LINKED ISSUE #396: Merge humanprotocol/fortune into humanprotocol/hmt-escrow
**Is your feature request related to a problem? Please describe.**
Merge fortune into hmt-escrow

**Describe the solution you'd like**

- Merge code
- Fix CI/CD
- Bring issues from fortune

portuu3:
Merge fortune repo into hmt-escrow


posix4e (changes requested):
Let's not duplicate the contracts directory. We can use a symlink if that makes your life easier

examples/fortune/docker-compose.yml
```diff
@@ -0,0 +1,104 @@
+version: '3.7'
```
posix4e, line 1: Can we use the docker-compose in the base directory

examples/fortune/reputation-oracle/contracts/EscrowAbi.json
```diff
@@ -0,0 +1,453 @@
+[
```
posix4e: Do not copy this file in, we should generate it

examples/fortune/exchange/src/contracts/EscrowFactoryAbi.json
```diff
@@ -0,0 +1,147 @@
+[
+{
```
posix4e: do not checkin, we should generate it

posix4e (changes requested):
Thanks for doing this

# TODO
- [x] Don't copy contracts, use a symlink if you must
- [x] Don't checkin the abi code, just generate them (we already generate them for the graph)
- [x] Rename the github action to be fortune based

examples/fortune/docker-compose.yml
portuu3: We can do it, but I wouldn't since it is just an example of use of hmt-escrow.
I have created a makefile to bring compile contracts, bring services up and deploy contracts with a single command like:
`make local`
Let me know if this is a good solution

PR #408, opened 2022-10-26
merge staking/slashing contracts

0xadev7:


posix4e (changes requested):
Overall looks very good and close. Some blocking next steps

1) We broke the subgraph tests
https://github.com/humanprotocol/hmt-escrow/actions/runs/3340625656/jobs/5530940776
image

2) How did you test this?
3) Are we doing any solidity lint?

contracts/EscrowFactory.sol
```diff
@@ -1,29 +1,58 @@
...
+    function createEscrow(address[] memory trustedHandlers)
+        public
+        returns (address)
+    {
+        require(staking != address(0), "Staking is not configured");
+        bool hasAvailalbeStake = IStaking(staking).hasAvailableStake(
+            msg.sender
+        );
+        require(
+            hasAvailalbeStake == true,
```
e271828-: typos

e271828- (commented):
Good start, but hard to review:
1. lint of existing files should be a separate PR
2. many contract functions / logic blocks missing comments. Need a comment for each with expected args + purpose so we can do initial review + write specs for formal verification of correctness.

PR #433, opened 2022-11-17
Add support download files drom public URI

alidzm:
In order to support run jobs that was deployed by anyone need to change code that downloads encrypted manifest from public URI.

Changes was applied for `def download`: if `key` contains http/https just upload resource using get request

### DIFF
--- a/hmt_escrow/storage.py
+++ b/hmt_escrow/storage.py
@@ -2,6 +2,8 @@ import hashlib
 import json
 import logging
 import os
+import urllib.request
+import re
 from typing import Dict, Tuple, Optional, Union
 
 import boto3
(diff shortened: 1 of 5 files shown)


Dzeranov (approved):

hmt_escrow/storage.py
```diff
@@ -163,7 +165,13 @@ def download(key: str, private_key: bytes, public: bool = False) -> Dict:
 
     """
     try:
-        content = download_from_storage(key=key, public=public)
+        url_pattern = "^https?:\\/\\/(?:www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-zA-Z0-9()]{1,6}\\b(?:[-a-zA-Z0-9()@:%_\\+.~#?&\\/=]*)$"
+        is_url = re.match(url_pattern, key)
+        content = (
```
igervasiychuk, line 170: why not just

if <condition>: 
  a
else:
 b

igervasiychuk: this Yoda-like construction is cool, but a bit unclear what's the benefit?

alidzm: I have another opinion, may be it's from JS like ternary operator (? :), but I can change the statement if you wish.

igervasiychuk: wish I not! Let it stay in.

